PR #7, opened 2021-09-08
unsigned corim cleanup

### LINKED ISSUE #3: Cleanup Unsigned CoRIM Container
Unsigned CoRIM has a profile field to set a specific implementation of CoRIM/CoMID using a Profile.

This github issue tracks the introduction of profile.

thomas-fossati:
Fixes #3

### DIFF
--- a/comid/measurement.go
+++ b/comid/measurement.go
@@ -316,10 +316,10 @@ func (o *Measurement) SetUEID(ueid eat.UEID) *Measurement {
 
 // SetUUID sets the supplied uuid in the measurement-values-map
 // of the target measurement
-func (o *Measurement) SetUUID(uuid UUID) *Measurement {
+func (o *Measurement) SetUUID(u UUID) *Measurement {
 	if o != nil {
 		// TODO(tho) check if uuid is valid
-		o.Val.UUID = &uuid
+		o.Val.UUID = &u
 	}
 	return o
 }
--- a/comid/measurement.go
+++ b/comid/measurement.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"net"
 
-	"github.com/google/uuid"
 	"github.com/veraison/eat"
 	"github.com/veraison/swid"
 )
@@ -189,8 +188,7 @@ func (o *Measurement) SetKeyUUID(u UUID) *Measurement {
 			return nil
 		}
 
-		// XXX(tho) this might be useless
-		if _, err := uuid.Parse(u.String()); err != nil {
+		if u.Valid() != nil {
 			return nil
 		}
 
@@ -256,12 +254,13 @@ func (o *Measurement) SetMinSVN(svn int64) *Measurement {
 // the target measurement
 func (o *Measurement) AddDigest(algID uint64, digest []byte) *Measurement {
 	if o != nil {
-		// TODO(tho) check len(digest) against algID
 		ds := o.Val.Digests
 		if ds == nil {
 			ds = NewDigests()
 		}
-		ds.AddDigest(algID, digest)
+		if ds.AddDigest(algID, digest) == nil {
+			return nil
+		}
 		o.Val.Digests = ds
 	}
 	return o
@@ -308,7 +307,9 @@ func (o *Measurement) SetSerialNumber(sn string) *Measurement {
 // of the target measurement
 func (o *Measurement) SetUEID(ueid eat.UEID) *Measurement {
 	if o != nil {
-		// TODO(tho) check if ueid is valid
+		if ueid.Validate() != nil {
+			return nil
+		}
 		o.Val.UEID = &ueid
 	}
 	return o
@@ -318,7 +319,9 @@ func (o *Measurement) SetUEID(ueid eat.UEID) *Measurement {
 // of the target measurement
 func (o *Measurement) SetUUID(u UUID) *Measurement {
 	if o != nil {
-		// TODO(tho) check if uuid is valid
+		if u.Valid() != nil {
+			return nil
+		}
 		o.Val.UUID = &u
 	}
 	return o
--- a/corim/unsignedcorim.go
+++ b/corim/unsignedcorim.go
@@ -130,6 +130,7 @@ func (o *UnsignedCorim) AddProfile(urlOrOID string) *UnsignedCorim {
 	return o
 }
 
+// Valid checks the validity (according to the spec) of the target unsigned CoRIM
 func (o UnsignedCorim) Valid() error {
 	if o.ID == (swid.TagID{}) {
 		return fmt.Errorf("empty id")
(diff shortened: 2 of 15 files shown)


comid/measurement.go
```diff
@@ -316,10 +316,10 @@ func (o *Measurement) SetUEID(ueid eat.UEID) *Measurement {
 
 // SetUUID sets the supplied uuid in the measurement-values-map
 // of the target measurement
-func (o *Measurement) SetUUID(uuid UUID) *Measurement {
+func (o *Measurement) SetUUID(u UUID) *Measurement {
 	if o != nil {
 		// TODO(tho) check if uuid is valid
-		o.Val.UUID = &uuid
+		o.Val.UUID = &u
```
yogeshbdeshpande, line 322: Would wonder, if you would like to correct the pending line 321 also, here, in case you are touching this function as part of tidy up??

corim/unsignedcorim.go
```diff
@@ -4,104 +4,219 @@
...
+		}
+
+		comidCBOR, err := c.ToCBOR()
+		if err != nil {
+			return nil
+		}
 
-	var u comid.UUID
+		// d9 01fa # tag(506)
+		comidTag := []byte{0xd9, 0x01, 0xfa}
```
yogeshbdeshpande, line 63: 506 is 0x01, 0xFA, could you please also add the description for D9 on line 62?

corim/unsignedcorim.go
```diff
@@ -4,104 +4,219 @@
...
+		taggedComid := append(comidTag, comidCBOR...)
 
-	return fmt.Errorf("unknown corim-id type (CBOR: %x)", data)
+		o.Tags = append(o.Tags, taggedComid)
+	}
+	return o
 }
 
-type Tag []byte
+// AddCoswid appends the CBOR encoded (and appropriately tagged) CoSWID to the
```
yogeshbdeshpande, line 72: Correct me if I am wrong but we should also have an Interface API to add a CBOR Encoded CoSWID directly into the CoRIM, rather than been passed as a structure as argument.

corim/unsignedcorim.go
thomas-fossati: d9 is the `tag()` in the description

corim/unsignedcorim.go
thomas-fossati: Sure, maybe we could add that in the future.  For now, I have preferred a tighter integration with the rest of the Veraison ecosystem.

yogeshbdeshpande: Future work is fine with me, so long as we track it via github issue. Also in Veraison PSA, perhaps we will have all the FW Measurements as Ref Values as part of one or more CoMID ?

corim/unsignedcorim.go
yogeshbdeshpande: Ah, I see!

corim/unsignedcorim.go
yogeshbdeshpande: Also thinking a bit more, if a CoRIM Creator is a different entity than a CoMID & CoSWID generators, then we should extend the same flexibility for CoMID also.

corim/unsignedcorim.go
```diff
@@ -4,104 +4,219 @@
...
 
-func (o *UnsignedCorim) SetIDString(s string) *UnsignedCorim {
+// AddDependentRim creates a corim-locator-map from the supplied arguments and
+// appends it to the dependent RIMs in the unsigned-corim-map
+func (o *UnsignedCorim) AddDependentRim(href string, thumbprint *swid.HashEntry) *UnsignedCorim {
 	if o != nil {
-		if s == "" {
-			return nil
+		l := Locator{
+			Href:       comid.TaggedURI(href),
```
yogeshbdeshpande, line 101: Just wondering reviewing line 101, does the dependent RIM href be translated to tagged URI from a CoMID package ? Does this look syntactically correct? Perhaps the same or similar API should exist in CoRIM package as well ?

corim/unsignedcorim.go
```diff
@@ -4,104 +4,219 @@
...
 	return o
 }
 
-func (o UnsignedCorim) GetIDString() (string, error) {
-	switch t := o.ID.val.(type) {
-	case string:
-		return t, nil
-	default:
-		return "", fmt.Errorf("corim-id type is: %T", t)
+func (o UnsignedCorim) Valid() error {
```
yogeshbdeshpande, line 133: It is not clear to me, what is the scope of Valid ? Could you please clarify this ?

corim/unsignedcorim.go
```diff
@@ -4,104 +4,219 @@
...
+	if o.ID == (swid.TagID{}) {
```
yogeshbdeshpande, line 134: I am thinking as a future enhancement, should we move TagID as a common package to be used both by CoSWID/SWID and CoMID packages ? Here checking o.ID == swid.TagID{} looks slightly odd to me ??

corim/unsignedcorim.go
```diff
@@ -4,104 +4,219 @@
...
 	"github.com/veraison/corim/comid"
 	"github.com/veraison/eat"
 	"github.com/veraison/swid"
 )
 
+// UnsignedCorim is the top-level representation of the unsigned-corim-map with
+// CBOR and JSON serialization.
 type UnsignedCorim struct {
-	ID            ID             `cbor:"0,keyasint" json:"corim-id"`
+	ID            swid.TagID     `cbor:"0,keyasint" json:"corim-id"`
```
yogeshbdeshpande, line 19: Similar comment as line 134, perhaps I should have commented that point here?

yogeshbdeshpande (commented):
LGTM, kindly clarify few of my points raised in the changes, then I am happy to approve!

comid/measurement.go
thomas-fossati: good point, see 5a9ac1f

corim/unsignedcorim.go
thomas-fossati: sure, see 898fe800

yogeshbdeshpande: Great Thanks for a quick revert! I will resolve the conversation now!

corim/unsignedcorim.go
thomas-fossati: I don't understand what is the oddity here: the check makes sure the tag is not empty.

yogeshbdeshpande: kindly check my oddity from https://github.com/veraison/corim/issues/9  perspective, what i am trying to say !

If you agree, we can resolve this conversation.

thomas-fossati: unfortunately I can't understand that either :-)

corim/unsignedcorim.go
thomas-fossati: yes, this might end up in a `corim/common` package

yogeshbdeshpande (approved):

PR #34, opened 2021-11-24
Set MKey as uint

yogeshbdeshpande:


comid/measurement.go
```diff
@@ -255,6 +276,17 @@ func (o *Measurement) SetKeyUUID(u UUID) *Measurement {
 	return o
 }
 
+// SetKeyUint sets the key of the target measurement-map to the supplied
+// uint
```
thomas-fossati, line 280:
```suggestion
// unsigned integer
```

comid/measurement_test.go
```diff
@@ -26,6 +26,14 @@ func TestMeasurement_NewUUIDMeasurement_empty_uuid(t *testing.T) {
 	assert.Nil(t, tv)
 }
 
+func TestMeasurement_NewUIntMeasurement_good_uint(t *testing.T) {
```
thomas-fossati, line 29: how can a uint be a bad uint?  What is this test testing?

comid/test_vars.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/veraison/eat"
 )
 
+var TestUInt uint = 45
```
thomas-fossati, line 18: why is this alone?  you can add it to the block below.

thomas-fossati (commented):
The code looks good, the testing needs to be moved from the examples to the units

comid/example_test.go
```diff
@@ -392,6 +499,134 @@ func Example_decode_JSON() {
 	// Output: OK
 }
 
+func Example_decode_JSON_Mkey_asUint() {
```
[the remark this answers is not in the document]
yogeshbdeshpande: Agree: will move this to Unit tests

comid/example_test.go
```diff
@@ -588,3 +823,48 @@ func Example_decode_CBOR_4() {
 
 	// Output: OK
 }
+
+func Example_decode_CBOR_5() {
```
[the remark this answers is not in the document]
yogeshbdeshpande: I will raise a github issue to track this, to add a suitable diag to this.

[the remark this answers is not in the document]
thomas-fossati: is this one of the examples from the corim-cddl repo like the other `Example_decode_CBOR_<n>()`?  If so, a URL into the repo is sufficient...

comid/measurement.go
```diff
@@ -19,7 +19,7 @@ type Measurement struct {
 }
 
 // Mkey stores a $measured-element-type-choice.
-// The supported types are UUID and PSA refval-id.
+// The supported types are UUID, PSA refval-id and unsigned integer
```
thomas-fossati, line 22: just a note: we should also support OID.

thomas-fossati: Now tracked in #35

comid/example_test.go
```diff
@@ -588,3 +588,49 @@ func Example_decode_CBOR_4() {
...
+	// TO DO Add suitable DIAG representing measurement with MKey as integer
+	in := []byte{
+		0xa5, 0x00, 0x65, 0x65, 0x6e, 0x2d, 0x47, 0x42, 0x01, 0xa1, 0x00, 0x78, 0x20, 0x6d, 0x79, 0x2d, 0x6e, 0x73, 0x3a, 0x61, 0x63, 0x6d, 0x65, 0x2d,
```
thomas-fossati, line 595: please fix the line limits consistently with the other similar test vectors

comid/example_test.go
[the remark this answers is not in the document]
yogeshbdeshpande: Done!

comid/example_test.go
```diff
@@ -590,38 +590,68 @@ func Example_decode_CBOR_4() {
 }
 
 func Example_decode_CBOR_5() {
-	// TO DO Add suitable DIAG representing measurement with MKey as integer
+	// Taken from https://github.com/ietf-corim-cddl/tree/main/examples
```
thomas-fossati, line 593: which one?  we need to be more precise here.

yogeshbdeshpande: Well there is no exact match now, as we are using a different Measurement Key.

As you said, I just gave a reference that the input material is taken from the examples given here!

yogeshbdeshpande: However, now done the change!

comid/measurement_test.go
```diff
@@ -112,3 +124,238 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
 
 	assert.Nil(t, tv.SetUUID(nonRFC4122UUID))
 }
+
+func Example_encode_uint_mkey() {
```
yogeshbdeshpande, line 128: Change the name to something else!

comid/measurement_test.go
```diff
@@ -129,6 +129,16 @@ var (
 	goldenMKey1 uint64 = 120
 )
 
+func TestUIntMkey_NotSet(t *testing.T) {
+	meas := &Measurement{}
```
thomas-fossati, line 133: it looks like `meas` is not needed, we could just use `mkey`

comid/measurement.go
```diff
@@ -153,6 +153,9 @@ func (o Mkey) MarshalJSON() ([]byte, error) {
 }
 
 func (o Mkey) MarshalCBOR() ([]byte, error) {
+	if err := o.Valid(); err != nil {
+		return nil, err
+	}
```
thomas-fossati, line 158: the validation check is usually done in the `ToCBOR` method - which, if the payload is deemed "valid", will call the custom marshaler.

comid/measurement_test.go
```diff
@@ -129,6 +129,16 @@ var (
...
+	mkey := &Mkey{}
+	meas.Key = mkey
+	expectedErr := "unknown measurement key type: <nil>"
+	err := meas.Key.Valid()
+	require.NotNil(t, err)
+	assert.EqualError(t, err, expectedErr)
+}
+
 func TestUIntMkey_ToCBOR_ok(t *testing.T) {
```
thomas-fossati, line 142: we don't seem to be testing the `ToCBOR` method here, why is the test called like that?

comid/measurement_test.go
```diff
@@ -155,7 +164,84 @@ func TestUIntMkey_ToCBOR_ok(t *testing.T) {
...
+			mkey:     0,
+			expected: "unknown measurement key type: <nil>",
+		},
+	}
+
+	for _, tv := range tvs {
+
+		mkey := &Mkey{}
+
+		require.NotNil(t, mkey)
```
thomas-fossati, line 184: this is unnecessary.  the allocation at line 182 can't possibly fail

comid/measurement_test.go
```diff
@@ -155,7 +164,84 @@ func TestUIntMkey_ToCBOR_ok(t *testing.T) {
 
 		assert.Nil(t, err)
 		assert.Equal(t, tv.expected, actual)
+	}
+}
+func TestUIntMkey_ToCBOR_notok(t *testing.T) {
```
thomas-fossati, line 169: the function needs changing: what we are testing is not the `ToCBOR` method

comid/measurement_test.go
```diff
@@ -155,7 +164,84 @@ func TestUIntMkey_ToCBOR_ok(t *testing.T) {
...
+
+	for _, tv := range tvs {
+
+		mkey := &Mkey{}
+
+		require.NotNil(t, mkey)
+
+		actual, err := mkey.MarshalCBOR()
+
+		fmt.Printf("JSON: %x\n", actual)
```
thomas-fossati, line 188: JSON?

comid/measurement_test.go
```diff
@@ -155,7 +164,84 @@ func TestUIntMkey_ToCBOR_ok(t *testing.T) {
...
+	}
+}
+func TestUIntMkey_ToCBOR_notok(t *testing.T) {
+	tvs := []struct {
+		mkey     uint64
+		expected string
+	}{
+		{
+			mkey:     0,
+			expected: "unknown measurement key type: <nil>",
```
thomas-fossati, line 176: this looks wrong.

comid/measurement_test.go
```diff
@@ -217,29 +328,28 @@ func TestUIntMkey_FromJSON_ok(t *testing.T) {
...
+func TestUIntMkey_FromJSON_notok(t *testing.T) {
 	tvs := []struct {
 		input    []byte
-		expected uint64
+		expected string
 	}{
 		{
-			input:    []byte{0x18, 0x64},
-			expected: goldenMKey,
+			input:    []byte("{\"type\":\"uint\",\"value\":\"abcdefg\"}"),
```
thomas-fossati, line 337: you can use back-ticks to encode a raw string (and thus avoid escaping).

comid/measurement_test.go
yogeshbdeshpande: No I verified, if we do not set the right key, it would still Marshal Blank CBOR and would not fail the test!

Hence I put the check for validity of mkey inside Marshal CBOR for mkey

comid/measurement_test.go
yogeshbdeshpande: Agree, verified and done now! thanks!

comid/measurement_test.go
yogeshbdeshpande: Corrected to use the right naming of Marshal and UnMarshal

comid/example_test.go
```diff
@@ -588,3 +588,79 @@ func Example_decode_CBOR_4() {
 
 	// Output: OK
 }
+
+func Example_decode_CBOR_5() {
+	// Taken from https://github.com/ietf-corim-cddl/blob/main/examples/comid-2.diag
```
thomas-fossati, line 593: we agreed this would go, at least until we add a new test vector into the ietf-corim-cddl repo that we can reference, right?

comid/measurement.go
```diff
@@ -37,8 +38,13 @@ func (o Mkey) Valid() error {
 		return nil
 	case TaggedPSARefValID:
 		return PSARefValID(t).Valid()
+	case uint64:
+		if o.val == nil {
+			return fmt.Errorf("empty uint64 Mkey")
```
thomas-fossati, line 43:
```suggestion
			return fmt.Errorf("empty uint Mkey")
```

comid/measurement.go
```diff
@@ -84,6 +99,15 @@ func (o *Mkey) UnmarshalJSON(data []byte) error {
 			)
 		}
 		o.val = TaggedPSARefValID(x)
+	case "uint":
+		var x uint64
+		if err := json.Unmarshal(v.Value, &x); err != nil {
+			return fmt.Errorf(
+				"cannot unmarshal $measured-element-type-choice of type uint64: %w",
```
thomas-fossati, line 106:
```suggestion
				"cannot unmarshal $measured-element-type-choice of type uint: %w",
```

comid/measurement.go
```diff
@@ -51,6 +57,15 @@ func (o Mkey) GetPSARefValID() (PSARefValID, error) {
 	}
 }
 
+func (o Mkey) GetKeyUint() (uint64, error) {
+	switch t := o.val.(type) {
+	case uint64:
+		return t, nil
+	default:
+		return 0, fmt.Errorf("measurement-key type is: %T", t)
```
thomas-fossati, line 65: Not that it really matters here, but only as a question of style I'd return `MAX_UINT64` rather than 0 here.

comid/measurement.go
```diff
@@ -92,7 +116,7 @@ func (o *Mkey) UnmarshalJSON(data []byte) error {
 }
 
 // MarshalJSON serializes the target Mkey into the type'n'value JSON object
-// uuid, psa.refval-id
+// uuid, psa.refval-id, unsigned integer
```
thomas-fossati, line 119:
```suggestion
// Supported types are: uuid, psa.refval-id and unsigned integer
```

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
 
 	assert.Nil(t, tv.SetUUID(nonRFC4122UUID))
 }
+
+var (
+	goldenMKey  uint64 = 100
+	goldenMKey1 uint64 = 120
+)
+
+func TestUIntMkey_NotSet(t *testing.T) {
```
thomas-fossati, line 132: Please fix the function name to go along the convention described above.

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+	mkey := &Mkey{}
+	expectedErr := "unknown measurement key type: <nil>"
+	err := mkey.Valid()
+	require.NotNil(t, err)
```
thomas-fossati, line 136: This is a test assertion (`assert`) not a prerequisite (`require`).  Anyway, it is also made redundant by the assertion on the following line, so it should be removed.

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+}
+
+func TestUIntMkey_MarshalCBOR_ok(t *testing.T) {
+	tvs := []struct {
+		mkey     uint64
+		expected []byte
+	}{
+		{
+			mkey:     goldenMKey,
+			expected: ([]byte{0x18, 0x64}),
```
thomas-fossati, line 147: use `MustHexDecode` here rather than using the raw values.

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+		},
+		{
+			mkey:     goldenMKey1,
+			expected: ([]byte{0x18, 0x78}),
```
thomas-fossati, line 151: ditto

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+		},
+	}
+
+	for _, tv := range tvs {
+		meas := NewUintMeasurement(tv.mkey)
+		require.NotNil(t, meas)
```
thomas-fossati, line 157: I'm not sure why we need to go through the full blown Measurement here?  Is it not sufficient to instantiate an MKey?

comid/measurement_test.go
thomas-fossati: What I meant is the logics here is wrong in the sense that we are not setting `tvs[i].mkey` anywhere.

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+	}
+}
+
+func TestUIntMkey_UnmarshalCBOR_ok(t *testing.T) {
+	tvs := []struct {
+		input    []byte
+		expected uint64
+	}{
+		{
+			input:    []byte{0x18, 0x64},
```
thomas-fossati, line 195: please use MustHexDecode

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+			expected: goldenMKey,
+		},
+		{
+			input:    []byte{0x18, 0x78},
```
thomas-fossati, line 199: ditto

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+			expected: goldenMKey1,
+		},
+	}
+
+	for _, tv := range tvs {
+		mKey := &Mkey{}
+		require.NotNil(t, mKey)
```
thomas-fossati, line 206: This is testing the obvious: the allocation cannot possibly fail

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+		},
+		{
+			input:    []byte(`{"type":"uint","value":123.456}`),
+			expected: "unexpected EOF",
+		},
+	}
+
+	for _, tv := range tvs {
+		mKey := &Mkey{}
+		require.NotNil(t, mKey)
```
thomas-fossati, line 233: ditto (redundant)

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+		expected string
+	}{
+		{
+			input:    []byte("Garbage"),
+			expected: "unexpected EOF",
+		},
+		{
+			input:    []byte(`{"type":"uint","value":123.456}`),
+			expected: "unexpected EOF",
+		},
```
thomas-fossati, line 228: since this is negative test for the CBOR decoder we should provide invalid CBOR here rather than invalid JSON, no?

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+	}
+
+	for _, tv := range tvs {
+		mKey := &Mkey{}
+		require.NotNil(t, mKey)
+
+		err := mKey.UnmarshalCBOR(tv.input)
+		assert.NotNil(t, err)
```
thomas-fossati, line 236: this is redundant, the following assert catches this condition as well

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+		},
+		{
+			mkey:     goldenMKey1,
+			expected: []byte(`{"type":"uint","value":120}`),
+		},
+	}
+
+	for _, tv := range tvs {
+
+		meas := NewUintMeasurement(tv.mkey)
```
thomas-fossati, line 259: same comment as in the positive CBOR encoder tests: we can directly look at the MKey, avoiding embedding it into a Measurement object.

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+			mkey:     0,
+			expected: "unknown measurement key type: <nil>",
+		},
+	}
+
+	for _, tv := range tvs {
+		mkey := &Mkey{}
+		actual, err := mkey.MarshalCBOR()
+
+		fmt.Printf("CBOR: %x\n", actual)
```
thomas-fossati, line 182: this may be garbage if err != nil in the line above

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+	for _, tv := range tvs {
+
+		meas := NewUintMeasurement(tv.mkey)
+
+		require.NotNil(t, meas)
+
+		actual, err := meas.Key.MarshalJSON()
+
+		fmt.Printf("JSON: %x\n", actual)
```
thomas-fossati, line 265: this may contain garbage if err != nil

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+		assert.Nil(t, err)
+		assert.Equal(t, tv.expected, actual)
```
thomas-fossati, line 268: you can use `assert.JSONEq`

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+func TestUIntMkey_ToJSON_notok(t *testing.T) {
+	tvs := []struct {
+		mkey     uint64
+		expected string
+	}{
+		{
+			mkey:     0,
+			expected: "unknown type <nil> for mkey",
+		},
+	}
```
thomas-fossati, line 281: this looks wrong for the same reason as its CBOR twin

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+	for _, tv := range tvs {
+
+		mkey := &Mkey{}
+
+		require.NotNil(t, mkey)
```
thomas-fossati, line 287: please remove this

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+		actual, err := mkey.MarshalJSON()
+
+		fmt.Printf("JSON: %x\n", actual)
```
thomas-fossati, line 291: at this point `actual` may contain garbage

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+		assert.NotNil(t, err)
```
thomas-fossati, line 293: redundant

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+		require.NotNil(t, mKey)
+
+		err := mKey.UnmarshalCBOR(tv.input)
+		assert.NotNil(t, err)
+
+		assert.EqualError(t, err, tv.expected)
+	}
+}
+
+func TestUIntMkey_ToJSON_ok(t *testing.T) {
```
thomas-fossati, line 242: we don't have a `ToJSON` method, do you mean `MarshalJSON` ?

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+		actual, err := meas.Key.MarshalJSON()
+
+		fmt.Printf("JSON: %x\n", actual)
+
+		assert.Nil(t, err)
+		assert.Equal(t, tv.expected, actual)
+	}
+}
+
+func TestUIntMkey_ToJSON_notok(t *testing.T) {
```
thomas-fossati, line 272: we don't have a `ToJSON` method, do you mean `MarshalJSON` ?

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+		},
+		{
+			input:    []byte(`{"type":"uint","value":120}`),
+			expected: goldenMKey1,
+		},
+	}
+
+	for _, tv := range tvs {
+		mKey := &Mkey{}
+		require.NotNil(t, mKey)
```
thomas-fossati, line 315: remove

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+		err := mKey.UnmarshalJSON(tv.input)
+		assert.Nil(t, err)
+		actual, err := mKey.GetKeyUint()
+		assert.Nil(t, err)
+		assert.Equal(t, tv.expected, actual)
+	}
+}
+
+func TestUIntMkey_FromJSON_notok(t *testing.T) {
```
thomas-fossati, line 325: We don't have a `FromJSON` method, do you mean `UnmarshalJSON` ?

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+		},
+		{
+			input:    []byte(`{"type":"uint","value":123.456}`),
+			expected: "cannot unmarshal $measured-element-type-choice of type uint64: json: cannot unmarshal number 123.456 into Go value of type uint64",
+		},
+	}
+
+	for _, tv := range tvs {
+		mKey := &Mkey{}
+		require.NotNil(t, mKey)
```
thomas-fossati, line 342: drop

comid/measurement_test.go
```diff
@@ -112,3 +123,227 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+		err := mKey.UnmarshalJSON(tv.input)
+		assert.NotNil(t, err)
```
thomas-fossati, line 345: drop

comid/test_vars.go
```diff
@@ -23,14 +23,15 @@ var (
...
-	TestTagID      = "urn:example:veraison"
+	TestOID               = "2.5.2.8192"
+	TestRegID             = "https://acme.example"
+	TestMACaddr, _        = net.ParseMAC("02:00:5e:10:00:00:00:01")
+	TestIPaddr            = net.ParseIP("2001:db8::68")
+	TestUEIDString        = "02deadbeefdead"
+	TestUEID              = eat.UEID(MustHexDecode(nil, TestUEIDString))
+	TestSignerID          = MustHexDecode(nil, "acbb11c7e4da217205523ce4ce1a245ae1a239ae3c6bfd9e7871f7e5d8bae86b")
+	TestTagID             = "urn:example:veraison"
+	TestUInt       uint64 = 45
```
thomas-fossati, line 34: is this used?

comid/measurement_test.go
yogeshbdeshpande: I thought about this. MKey does not live stand alone, so best use is to use the correct route to set Mkey which is via setting measurements, via newly created api.

comid/measurement_test.go
```diff
@@ -112,3 +121,216 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+		actual, err := meas.Key.MarshalJSON()
+		assert.Nil(t, err)
+		assert.Equal(t, tv.expected, actual)
+
+		fmt.Printf("JSON: %x\n", actual)
+	}
+}
+
+func TestMkey_MarshalJSON_uint_notok(t *testing.T) {
```
thomas-fossati, line 265: Same set of comments as above:
1. we have one only vector therefore we don't need a full-blown table driven test
but most importantly:
1. there is nothing specific to the uint mkey that is tested here

comid/measurement_test.go
```diff
@@ -112,3 +121,216 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+		err := mKey.UnmarshalJSON(tv.input)
+		assert.Nil(t, err)
+		actual, err := mKey.GetKeyUint()
+		assert.Nil(t, err)
+		assert.Equal(t, tv.expected, actual)
+	}
+}
+
+func TestMkey_UnmarshalJSON_uint_notok(t *testing.T) {
```
thomas-fossati, line 314: I'd expect the test vectors here to be such that GetKeyUint fails rather than the failure happening on the generic mkey unmarshaler

comid/measurement_test.go
```diff
@@ -112,3 +121,216 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+		err := mKey.UnmarshalCBOR(tv.mkey)
+		assert.Nil(t, err)
+		actual, err := mKey.GetKeyUint()
+		assert.Nil(t, err)
+		assert.Equal(t, tv.expected, actual)
+	}
+}
+
+func TestMkey_UnmarshalCBOR_uint_notok(t *testing.T) {
```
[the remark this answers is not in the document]
yogeshbdeshpande: Not clear what you are saying here!

In UnMarshal CBOR a Byte stream will be passed to the decoder. If an invalid data is passed, then 

it should fail Unmarshaling.

If the Unmarshal is a success, then GetKeyUint will never fail ?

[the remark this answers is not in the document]
yogeshbdeshpande: Anyway I think, I understood what you are trying to say, have corrected it now!

comid/measurement_test.go
yogeshbdeshpande: For 1. I agree, that I will perhaps add one more test so that it is justified to have a table driven logic.

comid/measurement_test.go
yogeshbdeshpande: I would prefer consistency and keep the code like this!

comid/measurement_test.go
```diff
@@ -112,3 +121,232 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+)
+
+func TestMkey_Valid_no_value(t *testing.T) {
+	mkey := &Mkey{}
+	expectedErr := "unknown measurement key type: <nil>"
+	err := mkey.Valid()
+	assert.EqualError(t, err, expectedErr)
+}
+
+func TestMeasurement_MarshalCBOR_uint_mkey_ok(t *testing.T) {
```
yogeshbdeshpande, line 137:
```suggestion
func TestMKey_MarshalCBOR_uint_mkey_ok(t *testing.T) {
```

comid/measurement_test.go
yogeshbdeshpande: &mkey{tv.mkey}

comid/measurement_test.go
```diff
@@ -112,3 +121,232 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+		err := mKey.UnmarshalCBOR(tv.mkey)
+		assert.Nil(t, err)
+		actual, err := mKey.GetKeyUint()
+		assert.Nil(t, err)
+		assert.Equal(t, tv.expected, actual)
+	}
+}
+
+func TestMkey_UnmarshalCBOR_uint_not_ok(t *testing.T) {
```
yogeshbdeshpande, line 219: Test for Mkey as Valid but Not a Uint Mkey

comid/measurement_test.go
yogeshbdeshpande: ok, change the test name to test mkey

comid/measurement_test.go
yogeshbdeshpande: Need to Implement this..

comid/measurement_test.go
```diff
@@ -112,3 +121,232 @@ func TestMeasurement_NewUUIDMeasurement_bad_uuid(t *testing.T) {
...
+
+		actual, err := meas.Key.MarshalJSON()
+		assert.Nil(t, err)
+		assert.Equal(t, tv.expected, actual)
+
+		fmt.Printf("JSON: %x\n", actual)
+	}
+}
+
+func TestMkey_MarshalJSON_uint_not_ok(t *testing.T) {
```
yogeshbdeshpande, line 275: Use the type: "Uint" and Key as NOT AN INTEGER THen see if the test fails

thomas-fossati:
Cool.  One last thing: it seems that we can increase the coverage threshold to 84.4% now!

yogeshbdeshpande:
Yes!

thomas-fossati:
> Yes!

can you move the needle in `.github/workflows/ci-go-cover.yml` please?

yogeshbdeshpande:
Sure will do!

yogeshbdeshpande:
Ambit in an another PR

PR #71, opened 2023-01-23
Update JWX dependency to v2

### LINKED ISSUE #70: CoRIM need to upgrade dependent  jwx package to a newer version
CoRIM uses jwx package.

https://lestrrat-go/jwx package needs to be upgraded from v.1.2.x to 
https://lestrrat-go/jwx/v2 v2.x to be compatible with veraison/services and other repositories.

thomas-fossati:
Fix #70

Signed-off-by: Thomas Fossati <thomas.fossati@arm.com>


corim/signer.go
```diff
@@ -4,35 +4,27 @@ import (
...
 	"fmt"
 	"reflect"
 
-	"github.com/lestrrat-go/jwx/jwk"
+	"github.com/lestrrat-go/jwx/v2/jwk"
 	cose "github.com/veraison/go-cose"
 )
 
-func getAlgAndKeyFromJWK(j string) (cose.Algorithm, crypto.Signer, error) {
+func getAlgAndKeyFromJWK(j []byte) (cose.Algorithm, crypto.Signer, error) {
```
yogeshbdeshpande, line 14: I did not changed the function but used the equivalent API ks.Key(0) which returns the key on Index 0 and it works, with minimal change!

thomas-fossati: since we don't deal in keysets it looks more correct to change this to use the right api

yogeshbdeshpande: However the earlier implementation was using key set, isn't it?

Anyway either way it is fine, we are using the same library tools!

yogeshbdeshpande:
Thanks for the quick change!

PR #86, opened 2023-08-29
Auth + config file support

setrofim:
Add support for remote service authentication to the `corim submit`
sub-command.

Associated with it is support for loading configuration from a file
(note: technically this was already present but was kind of broken and
wasn't actually used for any configuration.)

Make api-server configurable via the config file, as well as the command
line flag. This removes the need to specify --api-server flag on each
corim submit sub-command invocation (given that this setting is going to
be fairly static for a particular setup).


cocli/README.md
```diff
@@ -437,6 +437,16 @@ $ cocli corim submit \
 >> "corim.cbor" submit ok
 ```
 
+#### Remote Service Authentication
+
+The above will work if the remote service does not authenticate
```
yogeshbdeshpande, line 442: Just a question here, more than a comment.

Does the `cocli` configuration, in any way needs to be synchronised with the 
equivalent configuration on the services and apiclient side?
Specifically my question originates from the following line 11 on the config file:
 ( auth: none)

yogeshbdeshpande: Also it would be good to specify what is the default `cocli` behaviour!

cocli/data/config/example-config.yaml
```diff
@@ -0,0 +1,24 @@
+# This file contains example configuration for connecting to Veraison services.
+# This configuration is only necessary for the `cocli corim submit` sub-command,
+# as that is the only instance where remote service configuration is used. You
+# do not need this configuration for creating or monipulating corims and
```
yogeshbdeshpande, line 4:
```suggestion
# do not need this configuration for creating or manipulating corims and
```

yogeshbdeshpande (commented):
Left some comments, but in general LGTM!

yogeshbdeshpande (approved):
Minor nits but LGTM!

cocli/README.md
setrofim: This configuration is used to configure the `apiclient` so no synchronisation needed, as such, for that. The configuration for cocli does need to match the corresponding configuration for the service it is contacting.

The default is set in the example config, and is specified in the help for the corresponding flag (when you do `cocli corim submit  -h`).

PR #92, opened 2023-09-12
Partial v2 revert

setrofim:
Reverting module /v2 module name change. The reason is that go build system does not allow specifying a post-v1 module name unless there is a corresponding published release. Since we're not planning on publishing v2.0.0 until CoRIM spec is out of draft, we cannot rename the module until then. So for now, v2 development needs to happen on top of a v1 module name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,10 @@ SHELL := /bin/bash
 
 GO111MODULE := on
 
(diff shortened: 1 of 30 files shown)


thomas-fossati (approved):

yogeshbdeshpande (approved):
LGTM!

PR #93, opened 2023-09-14
Update comid templates for Crypto Keys

setrofim:
Update comid templates to use the correct format for keys, switching from the old representation to the new Crypto Key one.


yogeshbdeshpande (approved):
LGTM!

PR #96, opened 2023-11-03
Add CoRIM extension mechanisms

setrofim:
Add mechanisms to allow extending  `corim` implementation at the map-extension and type-choice extension points defined by the spec.


README.md
```diff
@@ -19,3 +19,16 @@ Before requesting a PR (and routinely during the dev/test cycle), you are encour
 make presubmit
 ```
 and check its output to make sure your code coverage figures are in line with the set target and that there are no newly introduced lint problems.
+
+## Extending CoRIM/CoMID
+
+The CoRIM specification provides for extensions via CDDL map-extension and
+type-choice sockets. `corim` and `comid` types that support map extensions
+embed and `Extensions` object that allows registering a struct defining
```
yogeshbdeshpande, line 27:
```suggestion
embed an `Extensions` object that allows registering a struct defining
```

extensions/README.md
```diff
@@ -0,0 +1,362 @@
+[CoRIM
+specification](https://www.ietf.org/archive/id/draft-ietf-rats-corim-02.htm)
+may be extended by other specifications at several points identified in the
```
yogeshbdeshpande, line 3:
```suggestion
may be extended by other specifications at well defined points identified in the
```

yogeshbdeshpande:
My apologies, my review is getting staggered. I will try to finish by tomorrow or by end of Friday .

README.md
```diff
@@ -19,3 +19,16 @@ Before requesting a PR (and routinely during the dev/test cycle), you are encour
...
 and check its output to make sure your code coverage figures are in line with the set target and that there are no newly introduced lint problems.
+
+## Extending CoRIM/CoMID
+
+The CoRIM specification provides for extensions via CDDL map-extension and
+type-choice sockets. `corim` and `comid` types that support map extensions
+embed an `Extensions` object that allows registering a struct defining
+extension fields. Type choice extensions can be implemented by calling an
+appropriate registration function and giving it a new type or (for enums)
+value.
```
yogeshbdeshpande, line 30:
```suggestion
The CoRIM specification provides a meachanism for adding extensions to the base CoRIM schema. The `corim` and `comid` maps which can be extended, embed an `Extensions` object  that allows registering a wrapper structure defining extension fields. For native types that can be extended, i.e. `type choice` extensions can be implemented by calling an appropriate registration function and giving it a new type or a value (for enums).
```

comid/classid.go
```diff
@@ -5,245 +5,375 @@ package comid
...
+//	  "value": <CLASS_ID_VALUE>
+//	}
+//
+// where <CLASS_ID_TYPE> must be one of the known IClassIDValue implementation
+// type names (available in this implementation: "uuid", "oid",
+// "psa.impl-id", "int"), and <CLASS_ID_VALUE> is the JSON encoding of the underlying
+// class id value. The exact encoding is <CLASS_ID_TYPE> dependent. For the base
+// implementation types it is
+//
+//	oid: dot-seprated integers, e.g. "1.2.3.4"
```
yogeshbdeshpande, line 92:
```suggestion
//	oid: dot-seperated integers, e.g. "1.2.3.4"
```

comid/classid.go
```diff
@@ -5,245 +5,375 @@ package comid
...
+func (o TaggedInt) Bytes() []byte {
+	var ret [8]byte
+	binary.BigEndian.PutUint64(ret[:], uint64(o))
+	return ret[:]
 }
 
-// Unset tests whether the target ClassID has been initialized
-func (o ClassID) Unset() bool {
-	return o.val == nil || o.Type() == ClassIDTypeUnknown
+type IClassIDFactory func(any) (*ClassID, error)
```
yogeshbdeshpande, line 349:
```suggestion
type IClassIDCreator func(any) (*ClassID, error)
```

yogeshbdeshpande, line 349: I would atleast add a statement, to line 349 like below:

// IClassIDCreator is an interface whose instance is provided by the Caller, when 
// registering a new ClassID Type

comid/classid.go
```diff
@@ -5,245 +5,375 @@ package comid
...
+var classIDValueRegister = map[string]IClassIDFactory{
+	OIDType:  NewOIDClassID,
+	UUIDType: NewUUIDClassID,
+	IntType:  NewIntClassID,
+
+	ImplIDType: NewImplIDClassID,
+}
+
+// RegisterClassIDType registers a new IClassIDValue implementation (created
+// by the provided IClassIDFactory) under the specified CBOR tag.
```
yogeshbdeshpande, line 360:
```suggestion
// by the provided IClassIDFactory) under the specified CBOR tag.
```
```suggestion
// by the provided IClassIDFactory) under the specified CBOR tag.
// The supplied IClassIDFactory must provide an implementation that handles
 // default `nil ` value
// nil value 
```

comid/classid.go
```diff
@@ -5,245 +5,375 @@ package comid
...
+func RegisterClassIDType(tag uint64, factory IClassIDFactory) error {
```
yogeshbdeshpande, line 361:
```suggestion
func RegisterClassIDType(tag uint64, newClassID IClassIDFactory) error {
```

comid/classid.go
setrofim: Why this change? "Factory" is the standard term for the pattern being implemented here.

comid/classid.go
setrofim: This would make it more confusing. The arguement is the the factory function, not a new class id. Something like `neewClassIDFunction` would make more sense, but it is more verbose, and again, "factory" is the standard well-understood term for the pattern being implemented here.

comid/classid.go
setrofim: (e.g. see https://en.wikipedia.org/wiki/Factory_method_pattern)

comid/classid.go
setrofim: This shoiuld be documented in the `IClasssIDFactory` interface documentation (which I need to add, as you've noted above), not here.

corim/extensions.go
```diff
@@ -0,0 +1,68 @@
...
+
+type ICorimValidator interface {
+	ValidCorim(*UnsignedCorim) error
+}
+
+type ISignerValidator interface {
+	ValidSigner(*Signer) error
+}
+
+type Extensions struct {
```
yogeshbdeshpande, line 21: This is extremely confusing to the reader.

You have same name spread everywhere.

Here inside `corim/extensions.go`
Then inside extensions/extensions.go
and same for comid/extensions.go

Perhaps you should consider, calling corim/extensions.go as

type CoRIMExtensions struct {
       extensions.Extensions
}

comid/extensions.go
```diff
@@ -0,0 +1,173 @@
...
+	if ok {
+		if err := ev.ValidTriples(triples); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (o *Extensions) ValidMval(triples *Mval) error {
```
yogeshbdeshpande, line 71: As discussed these methods should be private rather than private!

comid/cryptokey_test.go
```diff
@@ -399,3 +400,55 @@ func Test_NewCryptoKey_negative(t *testing.T) {
...
+
+func (o testCryptoKey) String() string {
+	return "test"
+}
+
+func (o testCryptoKey) Valid() error {
+	return nil
+}
+
+func Test_RegisterCryptoKeyTypeType(t *testing.T) {
```
yogeshbdeshpande, line 426:
```suggestion
func Test_RegisterCryptoKey(t *testing.T) {
```

comid/extensions.go
```diff
@@ -0,0 +1,173 @@
...
+	if ok {
+		if err := ev.ValidFlagsMap(triples); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (o *Extensions) SetTrue(flag Flag) {
```
yogeshbdeshpande, line 116: Please change them to local methods as well...

comid/svn.go
```diff
@@ -6,102 +6,254 @@ package comid
...
+	"strconv"
 
-type TaggedSVN uint64
-type TaggedMinSVN uint64
+	"github.com/veraison/corim/encoding"
+	"github.com/veraison/corim/extensions"
+)
 
+// SVN is the Security Version Number. This typically changes only when a
+// security relevant change is maded to the measured environment.
```
yogeshbdeshpande, line 16:
```suggestion
// security relevant change is needed to the measured environment.
```

comid/svn.go
```diff
@@ -6,102 +6,254 @@ package comid
...
 type SVN struct {
-	val interface{}
+	Value ISVNValue
 }
 
-func (o *SVN) SetSVN(val uint64) *SVN {
-	if o != nil {
-		o.val = TaggedSVN(val)
+// NewSVN creates a new SVN of the specified type with the specified value. The
```
yogeshbdeshpande, line 21:
```suggestion
// NewSVN creates a new SVN of the specified type and value. The
```

comid/svn.go
```diff
@@ -6,102 +6,254 @@ package comid
...
+// type must be one of the ones defined by the spec ("exact-value",
```
yogeshbdeshpande, line 22: ### 
```suggestion
// type must be one of the strings defined by the spec ("exact-value",
```

corim/meta.go
```diff
@@ -10,17 +10,31 @@ import (
...
 func NewSigner() *Signer {
 	return &Signer{}
 }
 
+// RegisterExtensions registers a struct as a collections of extensions
+func (o *Signer) RegisterExtensions(exts extensions.IExtensionsValue) {
+	o.Extensions.Register(exts)
+}
+
+// GetExtensions returns pervisouosly registered extension
```
yogeshbdeshpande, line 33:
```suggestion
// GetExtensions returns previously registered extension
```

corim/unsignedcorim.go
```diff
@@ -19,19 +20,39 @@ import (
...
 type UnsignedCorim struct {
-	ID            swid.TagID     `cbor:"0,keyasint" json:"corim-id"`
-	Tags          []Tag          `cbor:"1,keyasint" json:"tags"`
+	ID swid.TagID `cbor:"0,keyasint" json:"corim-id"`
+	// note: even though tags are mandatory for CoRIM, we allow omitting
+	// them in our JSON templates for cocli (the min template just has
+	// corim-id). Since we're never writing JSON (so far), this normally
+	// wouldn't matter, however the custom serialization code we use to
+	// handle embedded structs relies on the omitempty entry to determine
+	// if a filed is optional, so we use it during unmarshaling as well as
```
yogeshbdeshpande, line 29:
```suggestion
	// if a field is optional, so we use it during unmarshaling as well as
```

extensions/README.md
```diff
@@ -0,0 +1,362 @@
...
+   than types. This implementation provides registration functions for new
+   values for those types.
+
+> [!NOTE]
+> The CoRIM spec includes some CDDL from the CoSWID spec that also features
+> extension sockets. This code base does include the CoSWID implementation, and
+> so does not cover those extension points.
+
+
+## Map Extensions
```
yogeshbdeshpande, line 28: So Far we have been using the term `structures` for extending CoRIM Maps

extensions/README.md
```diff
@@ -0,0 +1,362 @@
...
+`Get()` returns an `interface{}`. There are equivalent `GetInt()`,
+`GetString()`, etc. methods that  perform the required conversions, and return
+the value of the indicated type, along with possible errors. ("Must" versions
+of these also exists, e.g. `MustGetString()`, that do not return an error).
+
+You can also get the pointer your extensions instance itself by calling
+extended type instance's `GetExtensions()`. This returns an `interface{}`, so
+you will need to type assert in order to be able to access the fields directly.
+
+### Introducing additional constraints
```
yogeshbdeshpande, line 72:
```suggestion
#### Introducing additional constraints
```

extensions/README.md
```diff
@@ -0,0 +1,362 @@
...
+	exts := entity.GetExtensions().(*EntityExtensions)
+	fmt.Printf("also entity email: %s\n", exts.Email)
+}
+```
+
+
+## Type Choice Extensions
+
+Type Choice extensions allow specifying alternative types for existing CoRIM
+fields by defining a struct that implements an appropriate interface and
```
yogeshbdeshpande, line 160:
```suggestion
fields by defining a  type that implements an appropriate interface and
```

yogeshbdeshpande: I am confused, I thought for Type extensions we will register the new type and a Callback which instantiates the new Type.

Why we are requesting users to create a structure?

extensions/README.md
setrofim: Theere are no structures in CoRIM, in the problem domain/concemptually, they are CoRIM map extensions, that are implemented suing `Extensions` `structs`. "Map" and "type choice" are the names of the extensions types being implemented.

extensions/README.md
setrofim: no three is correct -- it should be under "Map Extensions", at the same level as "adding new fields" as those are the two things you can do with map extensions.

extensions/README.md
setrofim: - "struct" rather than "type" is correct here, since we're talking about concrete implementations (the interface being implemented is the type).
  - I assume, by "callback" you mean "factory"? The factory still needs something to actually put into the new instance that that something is the struct that needs to be implemented. E.g. need to provide an implementation for both `IClassIDValue` which is what you're actually implementing, and a `IClassIDFactory` that produces a `*ClassID` that is populated with the underlying value of your type.

yogeshbdeshpande: > * "struct" rather than "type" is correct here, since we're talking about concrete implementations (the interface being implemented is the type).
> * I assume, by "callback" you mean "factory"? The factory still needs something to actually put into the new instance that that something is the struct that needs to be implemented. E.g. need to provide an implementation for both `IClassIDValue` which is what you're actually implementing, and a `IClassIDFactory` that produces a `*ClassID` that is populated with the underlying value of your type.

As discussed the user can define a type like taggedInt and provide the required methods for instantiation.

I think, we need to refine the language, to make it super clear to the consumer of the extensions!

extensions/README.md
```diff
@@ -0,0 +1,362 @@
...
+3. A couple of type-choice sockets (`$tag-rel-type-choice`,
+   `$corim-role-type-choice` and `$comid-role-type-choice`) define what, in
+   effect, are extensible enums. They allow providing additional values, rather
+   than types. This implementation provides registration functions for new
+   values for those types.
+
+> [!NOTE]
+> The CoRIM spec includes some CDDL from the CoSWID spec that also features
+> extension sockets. This code base does include the CoSWID implementation, and
+> so does not cover those extension points.
```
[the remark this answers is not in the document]
setrofim: Reworded.

extensions/README.md
```diff
@@ -0,0 +1,362 @@
...
+2. Some elements can contain values of one of pre-determined types. New types
+   can be provided by registering a factory function for those types using an
+   appropriate function. This (mostly) corresponds to the type-choice extension
+   sockets in the spec.
+3. A couple of type-choice sockets (`$tag-rel-type-choice`,
+   `$corim-role-type-choice` and `$comid-role-type-choice`) define what, in
+   effect, are extensible enums. They allow providing additional values, rather
+   than types. This implementation provides registration functions for new
+   values for those types.
+
```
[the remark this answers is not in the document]
setrofim: I have reworded to hopefully make thigns clearer, however I kept the numbed bullet formatting as this a list of the three extension mechanisms mentioned in the preceeding paragraph and ought to be formatted as such.

extensions/README.md
```diff
@@ -0,0 +1,373 @@
...
+> [!NOTE]
+> Some of the definitions in the CoRIM spec are "imported" from the CoSWID
+> spec.  Some of those definitions also feature extension sockets. However, as
+> they are defined in a different spec (and are not implemented in this
+> library), they cannot be extended using the above mechanisms.
+
```
yogeshbdeshpande, line 28:
```suggestion
> [!NOTE]
> CoRIM specification also "imports" CDDL from the CoSWID spec. Some of 
> these COSWID CDDL definitions also feature extension sockets. 

> However, as they are defined in a different spec and are implemented 
> in `veraison/coswid` library, they cannot be extended 
> using the extension feature provided in the CoRIM Library. The extension 
> support in CoRIM library is applicable ONLY to CoRIM  and CoMID Maps and 
> Types choices
```

yogeshbdeshpande: Request please review an attempt to clarify the Note point.

Feel free to edit my text to make it more clear, if you feel so!

extensions/README.md
```diff
@@ -0,0 +1,373 @@
...
+- `comid.Comid`
+- `comid.Entity`
+- `comid.FlagsMap`
+- `comid.Mval`
+- `comid.Triples`
+- `corim.Entity`
+- `corim.Signer`
+- `corim.UnsignedCorim`
+
```
yogeshbdeshpande, line 43: Please make note that the list will change as the specification evolves, this is applicable to `draft-02` of the specification

extensions/README.md
```diff
@@ -0,0 +1,373 @@
...
+being extended. This should be done as early as possible, before any marshaling
+is performed.
+
+These types can be extended in two ways: by adding additional fields, and by
+introducing additional constraints over existing fields.
+
+### Adding new fields
+
+To add a new fields, simply add them to your extensions struct, ensuring that
+the `cbor` and `json` tags on those fields are set corrects. As CoRIM spec
```
yogeshbdeshpande, line 56:
```suggestion
the `cbor` and `json` tags on those fields are set correctly. As CoRIM spec
```

extensions/README.md
```diff
@@ -0,0 +1,373 @@
...
+
+To introduce new constraints,  add a method called `Validate<type>(v *<type>)`
+to your extensions struct, where `<type>` is the name of the type being
+extended (one of the ones listed above) -- e.g.
+`ValidateComid(v *comid.Comid)` when extending `comid.Comid`. This method, if
+it exists, will be invoked inside the extended type instance's `Valid()`
+method, passing itself as the parameter.
+
+You do not need to define this method, unless you actually want to enforce some
+constraints (i.e. if you just want to define additional fields).
```
yogeshbdeshpande, line 83:
```suggestion
constraints.
```

setrofim: Why remove this clarification?

extensions/README.md
setrofim: There is a note in the top-level README to that effect. We should not be idetifying specific things that may change (because the implication then becomes that unless something is expressly indicated as subject to change, it won't change).

yogeshbdeshpande: Thanks, so long as we have note it is fine!

extensions/README.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Example
+
+The following example illustrates how to implement a map extension by extending `comid.Entity` with the following features:
+
+1. an optional "email" field
+2. additional validation to ensure that the existing name field contains a
```
yogeshbdeshpande, line 90: I have a clarification question here more than a comment.

What is the `use case` you are trying to highlight where the Validate method 
is checking for existing element validity ?

My concern is that a new user of extensions will get further confused on how the Validate is useful ?

For the benefit of simplicity can we keep it just checking the new parameters added for validity (like URI added is a valid URI?)

extensions/README.md
yogeshbdeshpande: The text is confusing, are you providing additional information in the braces `( )` ?

I found it to be counter intuitive to your original statement, which says !
Perhaps may be we simply say:
You do not need to define this method, if  additional constraints are not introduced by the new fields.

extensions/README.md
setrofim: You _can_ use Validate*  to  enforce individual field constraits on new fields, but it is not necessary for that, and it is not the reason it exists.  It is not necessary for that.

The things that the Valadiate* methods make possible that would not be possible other wise are:

- validating additional contraints on existing fields (since the types of those fields are fixed, you cannot do it the usual way, like you can with new fields)
- valdiating inter-field constraints (which, by their nature, cannot be captured in a type) 

The example illustrates that.

The primary mechanism for enforcing constraints on individual values is types. that's the sole reason for having types in the first place.

> OR I am setting an new Identifier which can be []byte but valid ONLY
when size is fixed to 32 bytes or 64 bytes.

You can do this by introducing your won type.  which is defined as "a seequence of exactly 32 or 64 bytes". 

If the example only showed validation of a new field, _that_ would be a sourc of confusion -- "why do I need to define Vadate method that take a whole other struct as an object, when I can just define my own type?"

extensions/README.md
setrofim: >if additional constraints are not introduced by the new fields.

??? the two things are orthogonal:

- new fields (do not require validation method -- constraints on new fields can be enforced via the fields' types)
- additional constraints (either on existing fields, or inter-field constraints)

The parentheciesd statement clarifies the situations where you might not want to add additional contraitas (when you're only adding new fields).

extensions/README.md
```diff
@@ -0,0 +1,376 @@
...
+`Extensions.Get()` passing in the name of the field you want to access. The
+name can be either the Go struct's field name, the name specified in the `json`
+tag, or (a string containing) the integer specified in the `cbor` tag.
+
+`Get()` returns an `interface{}`. There are equivalent `GetInt()`,
+`GetString()`, etc. methods that  perform the required conversions, and return
+the value of the indicated type, along with possible errors. ("Must" versions
+of these also exists, e.g. `MustGetString()`, that do not return an error).
+
+You can also get the pointer your extensions instance itself by calling
```
yogeshbdeshpande, line 72:
```suggestion
You can also get the pointer to your extensions instance itself by calling
```

extensions/README.md
yogeshbdeshpande: Thanks for all the descriptions. I think, a clarification is needed at the top to describe what this extension implies.

Then this explaination can have the right context

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+This implementation, likewise, allows dependent code to register extension
+types. This is done via three distinct extension mechanisms:
+
+1. Structures that allow extensions embed an `Extensions` object, which allows
+   registering a user-provided struct. The user-provided struct can extend
+   their containing structures in two ways:
+
+   - the fields of the user-provided struct become additional fields in
+     the containing structure.
+   - the user-provided struct may define addition constraints on the containing
```
yogeshbdeshpande, line 21:
```suggestion
   - the user-provided struct may define additional constraints on the containing
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+the value of the indicated type, along with possible errors. ("Must" versions
+of these also exists, e.g. `MustGetString()`, that do not return an error).
+
+You can also get the pointer your extensions instance itself by calling
+extended type instance's `GetExtensions()`. This returns an `interface{}`, so
+you will need to type assert in order to be able to access the fields directly.
+
+### Introducing additional constraints
+
+To introduce new constraints,  add a method called `Valid<TYPE>(v *<TYPE>)`
```
yogeshbdeshpande, line 93: I am thinking, really we are restricting and Constraining the type, here.

Should we name the method as 

`Constrain  <TYPE>' ?

yogeshbdeshpande: I mean ConstrainType

setrofim: done.

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+You do not need to define this method, unless you actually want to enforce some
+constraints (i.e. if you just want to define additional fields).
+
+### Example
+
+The following example illustrates how to implement a map extension by extending `comid.Entity` with the following features:
+
+1. an optional "email" field
+2. additional validation to ensure that the existing name field contains a
```
yogeshbdeshpande, line 108:
```suggestion
2. additional `constrain method` to ensure that the existing name field contains a
```

yogeshbdeshpande (approved):
LGTM!

Left Minor comment but in general good to submit from my PoV

comid/classid.go
```diff
@@ -5,245 +5,383 @@ package comid
...
-	return o.val == nil || o.Type() == ClassIDTypeUnknown
+func (o TaggedInt) Bytes() []byte {
+	var ret [8]byte
+	binary.BigEndian.PutUint64(ret[:], uint64(o))
+	return ret[:]
+}
+
+// IClassIDFactory defines the signature for the factory functions that may be
+// registred using RegisterClassIDType to provide a new implementation of the
+// corresponding type choice. The factory function should created a new *ClassID
```
thomas-fossati, line 351:
```suggestion
// corresponding type choice. The factory function should create a new *ClassID
```

comid/cryptokey.go
```diff
@@ -671,12 +703,61 @@ func NewCertPathThumbprint(he swid.HashEntry) (*CryptoKey, error) {
...
 	return key
 }
+
+func (o TaggedCertPathThumbprint) Type() string {
+	return CertPathThumbprintType
+}
+
+// ICryptoKeyFactory defines the signature for the factory functions that may be
+// registred using RegisterCryptoKeyType to provide a new implementation of the
+// corresponding type choice. The factory function should created a new *CryptoKey
```
thomas-fossati, line 722:
```suggestion
// corresponding type choice. The factory function should create a new *CryptoKey
```

comid/extensions.go
```diff
@@ -0,0 +1,173 @@
+// Copyright 2023 Contributors to the Veraison project.
```
thomas-fossati, line 1: I understand the overall (highly valuable) goal here, but it's not entirely clear to me what the criterion used for scoping the various "constrainers" types is.  I.e., are these all and only constrainers we are going to need?

(I guess there are going to be CoRIM constrainers as well, e.g., to check that a given CoRIM only contains CoMID tags, etc.)

comid/group.go
```diff
@@ -75,43 +60,97 @@ func (o *Group) UnmarshalCBOR(data []byte) error {
...
+	if err != nil {
+		panic(err)
+	}
+
+	return ret
+}
+
+// IGroupFactory defines the signature for the factory functions that may be
+// registred using RegisterGroupType to provide a new implementation of the
+// corresponding type choice. The factory function should created a new *Group
```
thomas-fossati, line 122: a pattern starts emerging :)

```suggestion
// corresponding type choice. The factory function should create a new *Group
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
+[CoRIM
+specification](https://www.ietf.org/archive/id/draft-ietf-rats-corim-02.htm)
```
thomas-fossati, line 2:
```suggestion
The [CoRIM
data format](https://datatracker.ietf.org/doc/draft-ietf-rats-corim)
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+may be extended by CoRIM Profiles documented in other specifications at well
+defined points identified in the base CoRIM spec using CDDL extension sockets.
+CoRIM profiles may
+
+1. Introduce new data fields to certain objects.
+2. Further constrain allowed values for existing fields.
+3. Introduce new type choices for fields who's values can be one of several
```
thomas-fossati, line 9:
```suggestion
3. Introduce new type choices for fields whose values can be one of several
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+   registering a user-provided struct. The user-provided struct can extend
+   their containing structures in two ways:
+
+   - the fields of the user-provided struct become additional fields in
+     the containing structure.
+   - the user-provided struct may define additional constraints on the
+     containing structure by defining an appropriate validation method for it.
+
+   This corresponds to the map-extension sockets in the spec.
+2. Some fields may have values one of a set of pre-determined types (e.g. an
```
thomas-fossati, line 25:
```suggestion
2. Some fields may have values chosen from a set of pre-determined types (e.g., an
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+> [!NOTE]
+> CoRIM specification also "imports" CDDL from the CoSWID spec. Some of
+> these COSWID CDDL definitions also feature extension sockets.
+
+> However, as they are defined in a different spec and are implemented
+> in `veraison/coswid` library, they cannot be extended
+> using the extension feature provided in the CoRIM Library. The extension
+> support in CoRIM library is applicable ONLY to CoRIM  and CoMID Maps and
+> Types choices
```
thomas-fossati, line 44:
```suggestion
> CoRIM also "imports" CDDL from the CoSWID spec. Some of
> these CoSWID CDDL definitions also feature extension sockets.
> However, as they are defined in a different spec and are implemented
> in the [`veraison/swid`](https://github.com/veraison/swid) library, they cannot be extended
> using the extension feature provided in the CoRIM library. The extension
> support in CoRIM library is applicable ONLY to CoRIM and CoMID maps and
> type choices
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+is performed.
+
+These types can be extended in two ways: by adding additional fields, and by
+introducing additional constraints over existing fields.
+
+### Adding new fields
+
+To add a new fields, simply add them to your extensions struct, ensuring that
+the `cbor` and `json` tags on those fields are set correctly. As CoRIM spec
+mandates integer keys, you must use `keyasint` option for the `cbor` tag.
```
thomas-fossati, line 75:
```suggestion
To add new fields, simply add them to your extensions struct, ensuring that
the `cbor` and `json` tags on those fields are set correctly. As CoRIM
mandates integer keys, you must use the `keyasint` option for the `cbor` tag.
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+To access the values of those fields, you can call the extended type instance's
+`Extensions.Get()` passing in the name of the field you want to access. The
+name can be either the Go struct's field name, the name specified in the `json`
+tag, or (a string containing) the integer specified in the `cbor` tag.
+
+`Get()` returns an `interface{}`. There are equivalent `GetInt()`,
+`GetString()`, etc. methods that  perform the required conversions, and return
+the value of the indicated type, along with possible errors. ("Must" versions
+of these also exists, e.g. `MustGetString()`, that do not return an error).
```
thomas-fossati, line 85:
```suggestion
To access the values of those fields, you can call the extended type instance's
`Extensions.Get()` passing in the name of the field you want to access. The
name can be either the Go struct's field name, the name specified in the `json`
tag, or (a string containing) the integer specified in the `cbor` tag.

`Get()` returns an `interface{}`. There are equivalent `GetInt()`,
`GetString()`, etc. methods that perform the required conversions, and return
the value of the indicated type, along with possible errors. ("Must" versions
of these also exist, e.g. `MustGetString()`, that do not return an error and instead call `panic()`).
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+You can also get the pointer your extensions instance itself by calling
+extended type instance's `GetExtensions()`. This returns an `interface{}`, so
+you will need to type assert in order to be able to access the fields directly.
```
thomas-fossati, line 89:
```suggestion
You can also get the pointer to your extension's instance itself by calling
the extended type instance's `GetExtensions()`. This returns an `interface{}`, so
you will need to type assert to be able to access the fields directly.
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+### Introducing additional constraints
+
+To introduce new constraints,  add a method called `Constrain<TYPE>(v *<TYPE>)`
+to your extensions struct, where `<TYPE>` is the name of the type being
+extended (one of the ones listed above) -- e.g.
+`ConstrainComid(v *comid.Comid)` when extending `comid.Comid`. This method, if
+it exists, will be invoked inside the extended type instance's `Valid()`
+method, passing itself as the parameter.
```
thomas-fossati, line 98:
```suggestion
To introduce new constraints, add a method called `Constrain<TYPE>(v *<TYPE>)`
to your extensions struct, where `<TYPE>` is the name of the type being
extended (one of the ones listed above) -- e.g.
`ConstrainComid(v *comid.Comid)` when extending `comid.Comid`. This method, if
it exists, will be invoked inside the extended type instance's `Valid()`
method, passing itself as the parameter.
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+You do not need to define this method, unless you actually want to enforce some
+constraints (i.e. if you just want to define additional fields).
```
thomas-fossati, line 101:
```suggestion
You do not need to define this method unless you actually want to enforce some
constraints (i.e., if you just want to define additional fields).
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+
+## Type Choice Extensions
+
+Type Choice extensions allow specifying alternative types for existing CoRIM
+fields by defining a type that implements an appropriate interface and
+registering it with CBOR tag.
+
+A type choice struct contains a single field, `Value`, that contains the actual
+object represented by the type choice. The `Value` implements an interface
+that is specific to the type choice and is derived from `ITypeChoiceValue`:
```
thomas-fossati, line 187:
```suggestion
Type Choice extensions allow specifying alternative types for existing CoRIM
fields by defining a type that implements an appropriate interface and
registering it with a CBOR tag.

A type choice struct contains a single field, `Value`, that contains the actual
object represented by the type choice. The `Value` implements an interface
that is specific to the type choice and is derived from `ITypeChoiceValue`:
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+- `comid.Mkey`
+- `comid.SVN`
+- `corim.EntityName`
+
+To provide a new value type, the following is required:
+
+1. Define a type that implements the value interface for the type choice you
+   want to extend. This interface is called `I<NAME>Value`,  where `<NAME>` is
+   the name of the type choice type(e.g. `IClassIDValue`).  These interfaces
+   always embed `ITypeChoiceValue` and possible define additional methods.
```
thomas-fossati, line 217:
```suggestion
1. Define a type that implements the value interface for the type choice you
   want to extend. This interface is called `I<NAME>Value`,  where `<NAME>` is
   the name of the type choice type(e.g. `IClassIDValue`).  These interfaces
   always embed `ITypeChoiceValue` and possibly define additional methods.
```

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+2. Create a factory function for your type, with the signature `func (any)
+   (*<NAME>, error)`, where `<NAME>` is the name of the type choice type that
+   will contain your value. (Note that the function must return a pointer to
+   the container type choice struct, _not_ to the value type you define.) This
+   function should create an instance of your value type from the provided
+   input and return a new type choice struct instance containing it. The range
+   of valid inputs is up to you, however it _must_ handle `null`, returning the
+   [zero-value](https://go.dev/ref/spec#The_zero_value) creating the zero value
+   for your type in that case.
```
thomas-fossati, line 226: _"[...] however it _must_ handle `null`, returning the [zero-value](https://go.dev/ref/spec#The_zero_value) creating the zero value for your type in that case."_

There seems to be some missing punctuation, or some amount of redundancy, which makes it difficult to understand.

extensions/README.md
```diff
@@ -0,0 +1,391 @@
...
+3. Register your factory function with the CBOR tag for your new type by
+   passing it to the registration function corresponding to the type choice
+   struct. It will have the name `Register<NAME>Type`, where `<NAME>` is the
+   name of the type choice struct that will contain your value (e.g.
+   `RegisterClassIDType`).
+
+### Example
+
+The following example illustrates how to add a new type choice value
+implementation by extending `CryptoKey` type to support DER values.
```
thomas-fossati, line 236:
```suggestion
implementation by extending the `CryptoKey` type to support DER values.
```

thomas-fossati (approved):
looks awesome, thanks for the titanic effort.

I have left a few editorial suggestions in line.

comid/extensions.go
setrofim: I.e., are these all and only constrainers we are going to need?

There is a constrainer interface corresponding to each map-extension point.  These are the only constrainers we can have, since contraints imposed by extensions can/should only be applied to the structures being extended.

(I guess there are going to be CoRIM constrainers as well, e.g., to check that a given CoRIM only contains CoMID tags, etc.)

There is already a constrainer for that in the `corim/corim` subpackage:

https://github.com/veraison/corim/blob/feature-extensions/corim/extensions.go#L13

comid/ccaplatformconfigid.go
```diff
@@ -3,11 +3,16 @@
...
-import "fmt"
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"unicode/utf8"
+)
 
-type CCAPlatformConfigID string
+var CCAPlatformConfigIDType = "cca.platform-config-id"
```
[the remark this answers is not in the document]
setrofim: I was just following the existing implementation here. I've beeing following the CoRIM spec and fixing things up where our implementation deviated, but I haven't checked that the PSA/CCA bits are aligned with the corresponding specs.

Given that this PR is already massive, maybe let leave this change to a separate PR; but yeah, we should probably this this.

extensions/README.md
setrofim: urgh, yeah, the "creating the zero value" bit is redundant.

extensions/README.md
setrofim: I'm specificicallying linking to draft 2 here since that is what we're following; The latest version may not match.

comid/ccaplatformconfigid.go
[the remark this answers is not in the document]
thomas-fossati: https://github.com/veraison/corim/issues/97

extensions/README.md
thomas-fossati: OK.  My comment was about using the IETF datatracker's which is supposed to have stronger URL stability.  If we need a precise version we use the versioned link https://datatracker.ietf.org/doc/draft-ietf-rats-corim/02

PR #98, opened 2023-12-12
Add CCA Realm Reference Values

yogeshbdeshpande:
CCA Realm Reference Value examples

### DIFF
--- a/cocli/data/templates/comid-cca-realm-refval.json
+++ b/cocli/data/templates/comid-cca-realm-refval.json
@@ -1,0 +1,70 @@
+{
+    "lang": "en-GB",
+    "tag-identity": {
+        "id": "43BBE37F-2E61-4B33-AED3-53CFF1428B16",
+        "version": 0
+    },
+    "entities": [
+        {
+            "name": "Workload Client Ltd.",
+            "regid": "https://workloadclient.example",
+            "roles": [
+                "tagCreator",
+                "creator",
+                "maintainer"
+            ]
+        }
+    ],
+    "triples": {
+        "reference-values": [
+            {
+                "environment": {
+                    "class": {
+                        "id": {
+                            "type": "uuid",
+                            "value": "CD1F0E55-26F9-460D-B9D8-F7FDE171787C"
+                        },
+                        "vendor": "Workload Client Ltd"
+                    }
+                },
+                "measurements": [
+                    {
+                        "value": {
+                            "digests": [
+                                "sha-384;QoS1aUymwNLPR4mguVrIAlyBjeUjBDZL580pgbLS7caFsyInfsJYGZYkE9jJssH1"
+                            ]
+                        }
+                    },
+                    {
+                        "value": {
+                            "digests": [
+                                "sha-384;IQe752H8pS2VE2oTVNt6TdV7Gya+DT2nHZ6yOYazS6YVq/ZRTPNeWp6lWgMtBop4"
+                            ]
+                        }
+                    },
+                    {
+                        "value": {
+                            "digests": [
+                                "sha-384;QQe752H8pS2VE2oTVNt6TdV7Gya+DT2nHZ6yOYazS6YVq/ZRTPNeWp6lWgMtBop4"
+                            ]
+                        }
+                    },
+                    {
+                        "value": {
+                            "digests": [
+                                "sha-384;JQe752H8pS2VE2oTVNt6TdV7Gya+DT2nHZ6yOYazS6YVq/ZRTPNeWp6lWgMtBop4"
+                            ]
+                        }
+                    },
+                    {
+                        "value": {
+                            "digests": [
+                                "sha-384;MQe752H8pS2VE2oTVNt6TdV7Gya+DT2nHZ6yOYazS6YVq/ZRTPNeWp6lWgMtBop4"
+                            ]
+                        }
+                    }
+                ]
+            }
+        ]
+    }
+}
--- a/cocli/data/templates/corim-cca-realm.json
+++ b/cocli/data/templates/corim-cca-realm.json
@@ -1,0 +1,25 @@
+{
+    "corim-id": "5c57e8f4-46cd-421b-91c9-08cf93e13cfc",
+    "dependent-rims": [
+        {
+            "href": "https://parent.example/rims/ccb3aa85-61b4-40f1-848e-02ad6e8a254b",
+            "thumbprint": "sha-256:5Fty9cDAtXLbTY06t+l/No/3TmI0eoJN7LZ6hOUiTXU="
+        }
+    ],
+    "profiles": [
+        "http://arm.com/cca/realm/1"
+    ],
+    "validity": {
+        "not-before": "2021-12-31T00:00:00Z",
+        "not-after": "2025-12-31T00:00:00Z"
+    },
+    "entities": [
+        {
+            "name": "ACME Ltd.",
+            "regid": "acme.example",
+            "roles": [
+                "manifestCreator"
+            ]
+        }
+    ]
+}
--- a/comid/example_cca_realm_refval_test.go
+++ b/comid/example_cca_realm_refval_test.go
@@ -1,0 +1,104 @@
+// Copyright 2023 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+
+package comid
+
+import "fmt"
+
+func Example_cca_realm_refval() {
+	comid := Comid{}
+
+	if err := comid.FromJSON([]byte(CCARealmRefValJSONTemplate)); err != nil {
+		panic(err)
+	}
+
+	if err := comid.Valid(); err != nil {
+		panic(err)
+	}
+
+	if err := extractRealmRefVals(&comid); err != nil {
+		panic(err)
+	}
+
+	// output:
+	// ClassID: cd1f0e5526f9460db9d8f7fde171787c
+	// Digest: 4284b5694ca6c0d2cf4789a0b95ac8025c818de52304364be7cd2981b2d2edc685b322277ec25819962413d8c9b2c1f5
+	// Digest: 2107bbe761fca52d95136a1354db7a4dd57b1b26be0d3da71d9eb23986b34ba615abf6514cf35e5a9ea55a032d068a78
+	// Digest: 4107bbe761fca52d95136a1354db7a4dd57b1b26be0d3da71d9eb23986b34ba615abf6514cf35e5a9ea55a032d068a78
+	// Digest: 2507bbe761fca52d95136a1354db7a4dd57b1b26be0d3da71d9eb23986b34ba615abf6514cf35e5a9ea55a032d068a78
+	// Digest: 3107bbe761fca52d95136a1354db7a4dd57b1b26be0d3da71d9eb23986b34ba615abf6514cf35e5a9ea55a032d068a78
+
+}
+
+func extractRealmRefVals(c *Comid) error {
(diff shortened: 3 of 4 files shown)


cocli/data/templates/comid-cca-realm-refval.json
```diff
@@ -0,0 +1,70 @@
...
+        }
+    ],
+    "triples": {
+        "reference-values": [
+            {
+                "environment": {
+                    "class": {
+                        "id": {
+                            "type": "uuid",
+                            "value": "CD1F0E55-26F9-460D-B9D8-F7FDE171787C"
```
thomas-fossati, line 25: It'd be handy to have a stable identifier for a given workload, but that's not the case in CCA.  I think we should use the RIM value to identify this (ephemeral) environment.

cocli/data/templates/comid-cca-realm-refval.json
```diff
@@ -0,0 +1,70 @@
...
+                            "digests": [
+                                "sha-384;JQe752H8pS2VE2oTVNt6TdV7Gya+DT2nHZ6yOYazS6YVq/ZRTPNeWp6lWgMtBop4"
+                            ]
+                        }
+                    },
+                    {
+                        "value": {
+                            "digests": [
+                                "sha-384;MQe752H8pS2VE2oTVNt6TdV7Gya+DT2nHZ6yOYazS6YVq/ZRTPNeWp6lWgMtBop4"
+                            ]
```
thomas-fossati, line 63: two observations:

1. CoRIM only allows one measurement-map in a reference-triple-record 
2. these measurements are not named, so it's not clear what they refer to (I assume there is some sort of positional convention to distinguish RIM from REMs?)

cocli/data/templates/comid-cca-realm-refval.json
```diff
@@ -0,0 +1,70 @@
...
+                        }
+                    }
+                ]
+            }
+        ]
+    }
+}
```
thomas-fossati, line 70: This looks very clumsy.  I suggest we extend the CoRIM measurement map with CCA-specific claims instead:

```json
[
  {
    "environment": {
      "instance": {
        "id": {
          "type": "arm-cca.rim",
          "value": "1000000000000000000000000000000000000000000000000000000000000000"
        }
      }
    }
  },
  {
    "measurement-map": {
      "mval": {
        "personalization": "5468652...",
        "RIM": "1000000000000000000000000000000000000000000000000000000000000000",
        "REM": [
          "0000000000000000000000000000000000000000000000000000000000000001",
          "0000000000000000000000000000000000000000000000000000000000000010",
          "0000000000000000000000000000000000000000000000000000000000000100",
          "0000000000000000000000000000000000000000000000000000000000010000"
        ],
        "RAK-hash-algo": "sha-256"
      }
    }
  }
]
```

cocli/data/templates/comid-cca-realm-refval.json
yogeshbdeshpande: Observation - 1: Correct: However we need to change that in the base CoRIM Schema.

All these Measurements belong to one `Environment` so really has to be in One Measurement Map

2. Yes, the positional convention is implicit, The first value will always be RIM Value and the subsequent will be REM's
3. We can add a comment in the `Supply Chain Documentation` that the first measurement always has to be `RIM`

cocli/data/templates/comid-cca-realm-refval.json
yogeshbdeshpande: RAK Hash Algorithm Identifier and Personalization Values will NOT come from Supply Chain!
They will not be provisioned.

Only provisioned will be a Workload Identity, RIM and REM.

yogeshbdeshpande: Also effectively RIM and REM are Digests and using Digest to represent Reference Values is perfectly correct!

yogeshbdeshpande: > Also effectively RIM and REM are Digests and using Digest to represent Reference Values is perfectly correct!
Instead of minting custom `map` with no added gain!

thomas-fossati: > RAK Hash Algorithm Identifier and Personalization Values will NOT come from Supply Chain!

why not?  And if not, should we expect another actor to supply those values?  If so, who?

cocli/data/templates/comid-cca-realm-refval.json
thomas-fossati: > Observation - 1: Correct: However we need to change that in the base CoRIM Schema.

That is not going to happen.  There is too much resistance to that.

cocli/data/templates/comid-cca-realm-refval.json
thomas-fossati: > We cannot request supply chain to ask for RIM to be the Identifier.

why not?

> It is perfectly fine to tag a `UUID` for the Stable Identifier. This is more practical!

How is it more practical to have to invent (and keep track of) an identifier that has no real linkage to the reference value data supplied?

yogeshbdeshpande: Supply Chain is free to choose any real identifier they want for their Workload.
The Class ID is extensible! 

This model proposes : UUID as a Workload Identifier. 

This ID will be relayed in the EAR(Veraison Extension) when RIM and REM Matches to `RP`,

In case two Workloads Have Identical RIM's we can then use the UUID to mention both of them in EAR.

cocli/data/templates/comid-cca-realm-refval.json
yogeshbdeshpande: I am sure about RPV it is set internally by the Hypervisor and if I understand correctly, external parties will not have visibility of the values. For RAK Hash Algorithm, I will check and update this thread!

cocli/data/templates/comid-cca-realm-refval.json
thomas-fossati: > Supply Chain is free to choose any real identifier they want for their Workload.
> The Class ID is extensible!
>
> This model proposes : UUID as a Workload Identifier.

I appreciate that.  My point is we are making things more complex than they need to be, without a good reason.

> This ID will be relayed in the EAR(Veraison Extension) when RIM and REM Matches to RP,

not sure, what is RP?

> In case two Workloads Have Identical RIM's we can then use the UUID to mention both of them in EAR.

I don't understand what "mention both of them in EAR" means.

yogeshbdeshpande: `RP` Here refers to Relying Party!
Mention both of them in EAT Attestation Results!

Also there is no complexity in tagging UUID as a Workload Identity. It is just an example Identifier to uniquely identify a Workload! 

Realm specs talks about GUID as one of the identifier for a Workload. Going by that logic: UUID makes perfect sense to me!

cocli/data/templates/comid-cca-realm-refval.json
thomas-fossati: > I am sure about RPV it is set internally by the Hypervisor and if I understand correctly, external parties will not have visibility of the values. 

no, both RPV and measurement algo are supplied as parameters at realm instantiation time to the VMM.

cocli/data/templates/comid-cca-realm-refval.json
yogeshbdeshpande: > > Observation - 1: Correct: However we need to change that in the base CoRIM Schema.
> 
> That is not going to happen. There is too much resistance to that.

I plan to re-initiate the discussion:

cocli/data/templates/comid-cca-realm-refval.json
yogeshbdeshpande: They are supplied at Realm Instantiation time to RMM by host, If I understand correctly!

cocli/data/templates/comid-cca-realm-refval.json
thomas-fossati: > Also there is no complexity in tagging UUID as a Workload Identity. It is just an example Identifier to uniquely identify a Workload!

I argue there *is* extra complexity because now the roles need to coordinate among themselves to keep track of yet another identifier which has no concrete relationship whatsoever with the workload.
 
> Realm specs talks about GUID as one of the identifier for a Workload.
> Going by that logic: UUID makes perfect sense to me!

Can you point me to the document where it is suggested to identify a given workload class using GUIDs?

cocli/data/templates/comid-cca-realm-refval.json
thomas-fossati: yep, so it's not set internally.  It's externally provided.

Anyway, I don't see any problem with making those optional (alongside REMs).

cocli/data/templates/comid-cca-realm-refval.json
yogeshbdeshpande: > > Also there is no complexity in tagging UUID as a Workload Identity. It is just an example Identifier to uniquely identify a Workload!
> 
> I argue there _is_ extra complexity because now the roles need to coordinate among themselves to keep track of yet another identifier which has no concrete relationship whatsoever with the workload.

Can you please explain what are the _Roles_ involved here: If  I understand correctly, there is ONLY one Supply Chain Entity the Work load Creator here!
> 
> > Realm specs talks about GUID as one of the identifier for a Workload.
> > Going by that logic: UUID makes perfect sense to me!
> 
> Can you point me to the document where it is suggested to identify a given workload class using GUIDs?

thomas-fossati: > This ID will be relayed in the EAR(Veraison Extension) when RIM and REM Matches to `RP`,

If that kind of payload identification is needed (on top of what is already provided by the trustworthiness vector), the relying party is going to want an endorsed value that provides information about the SW run by the workload rather than an opaque identifier.

cocli/data/templates/comid-cca-realm-refval.json
thomas-fossati: > I need to check again with our CCA experts before adding this as an Optional Parameter!
> 
> I am happy to do that, if I find that there is a possibility `Hypervisor` Gets the same from external sources:

```shell
 $ lkvm run --realm 				 \
	 --measurement-algo=["sha256", "sha512"] \
	 --disable-sve				 \
	 <normal-vm-options>
```

😄

yogeshbdeshpande: the measurement-algo here may be the hash algorithm used for Calculating RIM and REM Measurements? Isn't it?

thomas-fossati: *all* realm measurmements (see §A7.2.3.1.5 of the realm spec.)

cocli/data/templates/comid-cca-realm-refval.json
thomas-fossati: > > I argue there _is_ extra complexity because now the roles need to coordinate among themselves to keep track of yet another identifier which has no concrete relationship whatsoever with the workload.
> 
> Can you please explain what are the _Roles_ involved here: If I understand correctly, there is ONLY one Supply Chain Entity the Work load Creator here!

RP and RVP.

cocli/data/templates/comid-cca-realm-refval.json
yogeshbdeshpande: yes this `measurement-algo` is expected to be the one set by External Supply Chain.

It is different then RAK Hash Algo, which was referred earlier!

PR #107, opened 2024-01-30
Add Integrity registers

### LINKED ISSUE #108: Implement Integrity Register changes
Some environments (e.g., Arm CCA Realms, TPMs) have a bunch of measurements bundled in Register sets (like PCR Banks) rather than single digests.

The full requirement was captured in https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/185
while the associated change in the CoRIM Specification was merged via: 
https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/189

This issue tracks the `veraison/corim` implementation of the above requirement.

yogeshbdeshpande:
This PR implements the recently added Integrity Registers (https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#name-integrity-registers) to the CoRIM schema.

Fixes #108

### DIFF
--- a/comid/integrityregisters.go
+++ b/comid/integrityregisters.go
@@ -1,0 +1,138 @@
+// Copyright 2024 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+package comid
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"github.com/veraison/swid"
+)
+
+const TextType = "text"
+
+// IRegisterIndex is the interface to hold register index
+// Supported index types are uint and text
+type IRegisterIndex interface{}
+
+// IntegrityRegisters holds the Integrity Registers
+type IntegrityRegisters struct {
+	m map[IRegisterIndex]Digests
+}
+
+func NewIntegrityRegisters() *IntegrityRegisters {
+	return &IntegrityRegisters{m: make(map[IRegisterIndex]Digests)}
+}
+
+// AddDigests allows inserting an array of digests at a specific index
+// Supported index types are uint and text
+func (i *IntegrityRegisters) AddDigests(index IRegisterIndex, digests Digests) error {
+	if len(digests) == 0 {
+		return fmt.Errorf("no digests to add")
+	}
+	for _, digest := range digests {
+		if err := i.AddDigest(index, digest); err != nil {
+			return fmt.Errorf("unable to add Digest: %w", err)
+		}
+	}
+	return nil
+}
+
+// AddDigest allows inserting a digest at a specific index
+// Supported index types are uint and text
+func (i *IntegrityRegisters) AddDigest(index IRegisterIndex, digest swid.HashEntry) error {
+	if i.m == nil {
+		return fmt.Errorf("no register to add digest")
+	}
+	switch t := index.(type) {
+	case string, uint, uint64:
+		i.m[t] = append(i.m[t], digest)
+	default:
+		return fmt.Errorf("unexpected type for index: %T", t)
+	}
+	return nil
+}
+
+func (i IntegrityRegisters) MarshalCBOR() ([]byte, error) {
+	return em.Marshal(i.m)
+}
+
+func (i *IntegrityRegisters) UnmarshalCBOR(data []byte) error {
+	return dm.Unmarshal(data, &i.m)
+}
+
+type keyTypeandVal struct {
+	KeyType string `json:"key_type"`
+	Value   json.RawMessage
+}
+
+func (i IntegrityRegisters) MarshalJSON() ([]byte, error) {
+	jmap := make(map[string]json.RawMessage)
+	var newkey string
+	for key, val := range i.m {
+		var ktv keyTypeandVal
+		switch t := key.(type) {
+		case uint, uint64:
+			ktv.KeyType = UintType
+			newkey = fmt.Sprintf("%v", key)
+		case string:
+			ktv.KeyType = TextType
+			newkey = key.(string)
+		default:
+			return nil, fmt.Errorf("unknown type %T for index-type-choice", t)
+		}
+
+		newval, err := json.Marshal(val)
+		if err != nil {
+			return nil, err
+		}
+		ktv.Value = newval
+		Value, err := json.Marshal(ktv)
+		if err != nil {
+			return nil, err
+		}
+		jmap[newkey] = Value
+	}
+	return json.Marshal(jmap)
+}
+
+func (i *IntegrityRegisters) UnmarshalJSON(data []byte) error {
+	if i.m == nil {
+		i.m = make(map[IRegisterIndex]Digests)
+	}
+	jmap := make(map[string]json.RawMessage)
+	var index IRegisterIndex
+	if err := json.Unmarshal(data, &jmap); err != nil {
+		return fmt.Errorf("register map decoding failure: %w", err)
+	}
+	for key, val := range jmap {
+		var ktv keyTypeandVal
+		var d Digests
+
+		if err := json.Unmarshal(val, &ktv); err != nil {
+			return fmt.Errorf("unable to unmarshal keyTypeAndValue: %w", err)
+		}
+		if err := json.Unmarshal(ktv.Value, &d); err != nil {
+			return fmt.Errorf("unable to unmarshal Digests: %w", err)
+		}
+		switch ktv.KeyType {
+		case UintType:
+			u, err := strconv.Atoi(key)
+			if err != nil {
+				return fmt.Errorf("unable to convert key to uint: %w", err)
+			} else if u < 0 {
+				return fmt.Errorf("invalid negative integer key")
+			}
+			index = uint(u)
+		case TextType:
+			index = key
+		default:
+			return fmt.Errorf("unexpected key type for index: %s", ktv.KeyType)
+		}
+		if err := i.AddDigests(index, d); err != nil {
+			return fmt.Errorf("unable to add digests into register set: %w", err)
+		}
+	}
+	return nil
+}
--- a/comid/integrityregisters_test.go
+++ b/comid/integrityregisters_test.go
@@ -1,0 +1,302 @@
+// Copyright 2024 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+
+package comid
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/veraison/swid"
+)
+
+func prepareRegister(t *testing.T, iType string) (*IntegrityRegisters, error) {
+	var err error
+	val := MustHexDecode(t, "e45b72f5c0c0b572db4d8d3ab7e97f368ff74e62347a824decb67a84e5224d75")
+	entry := swid.HashEntry{HashAlgID: swid.Sha256, HashValue: val}
+	reg := NewIntegrityRegisters()
+	for index := 0; index < 5; index++ {
+		switch iType {
+		case "uint":
+			err = reg.AddDigest(uint(index), entry)
+			require.NoError(t, err)
+		case "text":
+			i := fmt.Sprint(index)
+			err = reg.AddDigest(i, entry)
+			require.NoError(t, err)
+		default:
+			err = fmt.Errorf("invalid iType = %s", iType)
+		}
+		if err != nil {
+			return nil, err
+		}
+	}
+	return reg, nil
+}
+
+func TestIntegrityRegisters_AddDigest_OK(t *testing.T) {
+	_, err := prepareRegister(t, "uint")
+	require.NoError(t, err)
+}
+
+func TestIntegrityRegisters_AddDigest_NOK(t *testing.T) {
+	expectedErr := `no register to add digest`
+	register := IntegrityRegisters{}
+	err := register.AddDigest(uint(0), swid.HashEntry{})
+	assert.EqualError(t, err, expectedErr)
+	expectedErr = `unexpected type for index: bool`
+	var k bool
+	reg, err := prepareRegister(t, "uint")
+	require.NoError(t, err)
+	err = reg.AddDigest(k, swid.HashEntry{})
+	assert.EqualError(t, err, expectedErr)
+}
+
+func TestIntegrityRegisters_AddDigests_NOK(t *testing.T) {
+	expectedErr := `no digests to add`
+	register := IntegrityRegisters{}
+	err := register.AddDigests(uint(0), []swid.HashEntry{})
+	assert.EqualError(t, err, expectedErr)
+}
+
+func TestIntegrityRegisters_MarshalCBOR_UIntIndex_OK(t *testing.T) {
+	reg, err := prepareRegister(t, "uint")
+	// Below is the partial CBOR Pretty notation to highlight index as uint: unsigned(0)
+
+	// A5                                      # map(5)
+	//    00                                   # unsigned(0)
+	//    81                                   # array(1)
+	//       82                                # array(2)
+	//          01                             # unsigned(1)
+	//          58 20                          # bytes(32)
+	expected := MustHexDecode(t, "a5008182015820e45b72f5c0c0b572db4d8d3ab7e97f368ff74e62347a824decb67a84e5224d75018182015820e45b72f5c0c0b572db4d8d3ab7e97f368ff74e62347a824decb67a84e5224d75028182015820e45b72f5c0c0b572db4d8d3ab7e97f368ff74e62347a824decb67a84e5224d75038182015820e45b72f5c0c0b572db4d8d3ab7e97f368ff74e62347a824decb67a84e5224d75048182015820e45b72f5c0c0b572db4d8d3ab7e97f368ff74e62347a824decb67a84e5224d75")
+	require.NoError(t, err)
+	actual, err := reg.MarshalCBOR()
+	require.NoError(t, err)
+	fmt.Printf("CBOR Payload = %x\n", actual)
+	assert.Equal(t, expected, actual)
+}
+
+func TestIntegrityRegisters_UnmarshalCBOR_UIntIndex_OK(t *testing.T) {
+	expected := IntegrityRegisters{map[IRegisterIndex]Digests{
+		uint64(0): []swid.HashEntry{{HashAlgID: swid.Sha256, HashValue: MustHexDecode(t, "e45b72f5c0c0b572db4d8d3ab7e97f368ff74e62347a824decb67a84e5224d75")}},
+		uint64(1): []swid.HashEntry{{HashAlgID: swid.Sha256, HashValue: MustHexDecode(t, "e45b72f5c0c0b572db4d8d3ab7e97f368ff74e62347a824decb67a84e5224d75")}},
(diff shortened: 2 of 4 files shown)


comid/integregisters.go
```diff
@@ -0,0 +1,152 @@
...
+	"fmt"
+
+	"github.com/veraison/swid"
+)
+
+// IRegisterIndex is the interface to hold register index
+type IRegisterIndex interface{}
+
+// IntegReg holds the Integrity Registers
+type IntegReg struct {
```
thomas-fossati, line 16: The choices in naming are very confusing: there are 5 different contractions for the same concept:
* IntegRegister
* IntegReg
* integregisters
* IRegisterIndex
* Register (which could be just `m` (map) or `v` (val) or `val`)

There is no good reason for shortening `IntegrityRegister`.  On the contrary, there are very good reasons to make sure the code can be related unambiguously to the spec.

Let's use `IntegrityRegister` consistently (also the file names).

comid/integregisters.go
```diff
@@ -0,0 +1,152 @@
...
+			return fmt.Errorf("unable to UnMarshal Digests: %w", err)
+		}
+		if err = i.AddDigests(index, d); err != nil {
+			return fmt.Errorf("unable to insert register into register set: %w", err)
+		}
+	}
+	return nil
+}
+
+func unMarshalIndex(key string) (IRegisterIndex, error) {
```
thomas-fossati, line 131: unmarshal is a single word, camel case does not apply:

```suggestion
func unmarshalIndex(key string) (IRegisterIndex, error) {
```

comid/integregisters_test.go
```diff
@@ -0,0 +1,230 @@
...
+	for index, digests := range reg.Register {
+		_, ok := index.(uint)
+		assert.True(t, ok)
+		err = digests.Valid()
+		require.NoError(t, err)
+	}
+}
+
+func TestIntegRegister_FromJSON_TextUInt_Index_OK(t *testing.T) {
+	j := `{
```
thomas-fossati, line 170: why do we need the escape?  It makes the test vector hard to read.

comid/integregisters.go
```diff
@@ -0,0 +1,152 @@
...
+
+// IntegReg holds the Integrity Registers
+type IntegReg struct {
+	Register map[IRegisterIndex]Digests
+}
+
+func NewIntegRegister() (*IntegReg, error) {
+	reg := make(map[IRegisterIndex]Digests)
+	return &IntegReg{Register: reg}, nil
+}
```
thomas-fossati, line 23: this cannot fail, returning `error` is superfluous.

comid/integregisters_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/veraison/swid"
+)
+
+func prepareRegister(iType string) (*IntegReg, error) {
+	tv := "12435465869790801243546586979080124354658697908"
```
thomas-fossati, line 16: It's quite weird to have the sha384 output as a string.  suggest using MustHexDecode("...").  But more importantly, why do we need to vary the entries?  Are they checked somewhere?  From a quick scan, it doesn't seem to be the case.

comid/integregisters_test.go
```diff
@@ -0,0 +1,230 @@
...
+	reg, err := NewIntegRegister()
+	if err != nil {
+		return nil, err
+	}
+
+	for index := 0; index < 5; index++ {
+		entry := &swid.HashEntry{}
```
thomas-fossati, line 23: is there any reason for `entry` to be a pointer?

comid/integregisters_test.go
```diff
@@ -0,0 +1,230 @@
...
+		d := tv + fmt.Sprint(index)
+		fmt.Printf("test bytes = %s", tv)
```
thomas-fossati, line 25: spurious line

comid/integregisters.go
```diff
@@ -0,0 +1,152 @@
...
+	return nil
+}
+
+func (i IntegReg) MarshalCBOR() ([]byte, error) {
+	return em.Marshal(i.Register)
+}
+
+func (i *IntegReg) UnMarshalCBOR(data []byte) error {
+	return dm.Unmarshal(data, &i.Register)
+}
```
thomas-fossati, line 73: why are the marshaling and unmarshaling APIs for CBOR and JSON not symmetric?

comid/integregisters.go
```diff
@@ -0,0 +1,152 @@
...
+	case string, uint, uint64:
+		i.Register[t] = append(i.Register[t], digest)
+	default:
+		return fmt.Errorf("unexpected type for index: %T", t)
+	}
+	return nil
+}
+
+// RemoveIndex removes an Index from the Register
+func (i *IntegReg) RemoveIndex(index IRegisterIndex) error {
```
thomas-fossati, line 54: This doesn't seem to provide any useful functionality.

yogeshbdeshpande:
@thomas-fossati Thank you so much for the comments! I shall review the same and discuss if required.

thomas-fossati:
> @thomas-fossati Thank you so much for the comments! I shall review the same and discuss if required.

👍 

One more thing: could you please squash the two commits into one?  There seems to be no reason for their separate existence.

yogeshbdeshpande:
> > @thomas-fossati Thank you so much for the comments! I shall review the same and discuss if required.
> 
> 👍
> 
> One more thing: could you please squash the two commits into one? There seems to be no reason for their separate existence.

Indeed!

yogeshbdeshpande:
> > > @thomas-fossati Thank you so much for the comments! I shall review the same and discuss if required.
> > 
> > 
> > 👍
> > One more thing: could you please squash the two commits into one? There seems to be no reason for their separate existence.
> 
> Indeed!

I have done the squash and will incorporate the comments, as discussed!

comid/integregisters.go
```diff
@@ -0,0 +1,152 @@
...
+
+func (i IntegReg) MarshalCBOR() ([]byte, error) {
+	return em.Marshal(i.Register)
+}
+
+func (i *IntegReg) UnMarshalCBOR(data []byte) error {
+	return dm.Unmarshal(data, &i.Register)
+}
+
+func (i IntegReg) ToJSON() ([]byte, error) {
```
setrofim, line 75: If I understand correctly, this mashals the key as a tnv JSON object, which is
then marshalled as string so that it can be used as a JSON key?
E.g.

```json
{
	"{\"type\": \"uint\", \"value\": 7}": [... encoded digests ...]
}
```

This double marshalling is weird and will resuilt in JSON text that is
difficult to prase visually. Why not move the key type to be part of the value,
so if would look something like

```json
{
	"7": {
		"key_type": "uint",
		"value": [... encoded digests ...]
	}
}
```

yogeshbdeshpande: Hmmm, I am not sure this representation accurately describes the `key type` though!

If I am not mis-taken: Here you are representing the `key`  as "7" in JSON which is a string representation

setrofim: To illustrate,

```json
{
	"7": {
		"key_type": "uint",
                [ ... ]
	}
}
```
Means the key is `7` (integer) 
```json
{
	"7": {
		"key_type": "text",
                [ ... ]
	}
}
```
means the key is `"7"` (string).

yogeshbdeshpande: Let me check, if I can improve the existing test-vector, that provides the required simplification without losing the `homogeneity`  offered by the `tnv` pattern we have been using consistently throughout this repository!

comid/integregisters.go
yogeshbdeshpande: Renamed everywhere to IntegrityRegisters

comid/integregisters.go
yogeshbdeshpande: Removed it for now!

comid/integregisters.go
setrofim: In either case, you're breaking away form the `tnv` pattern, as it only works for JSON values -- that encodes a value a JSON object, which cannot be a key. In the original approach, you're then encoding the  `tnv` JSON object as a string, so it's no longer a `tnv` object.

yogeshbdeshpande: Done as required, other alternatives did not render very cleanly so implemented as above!

comid/integrityregisters.go
```diff
@@ -0,0 +1,134 @@
...
+type IRegisterIndex interface{}
+
+// IntegrityRegisters holds the Integrity Registers
+type IntegrityRegisters struct {
+	Register map[IRegisterIndex]Digests
+}
+
+func NewIntegrityRegisters() *IntegrityRegisters {
+	reg := make(map[IRegisterIndex]Digests)
+	return &IntegrityRegisters{Register: reg}
```
thomas-fossati, line 25: simpler and more idiomatically
```suggestion
	return &IntegrityRegisters{
		make(map[IRegisterIndex]Digests),
	}
```

comid/integrityregisters.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+const TextType = "text"
+
+// IRegisterIndex is the interface to hold register index
+type IRegisterIndex interface{}
+
+// IntegrityRegisters holds the Integrity Registers
+type IntegrityRegisters struct {
+	Register map[IRegisterIndex]Digests
```
thomas-fossati, line 20: `Register` doesn't need to be public.  (Hint: in VSC you can just "Rename Symbol" to some private name - e.g., `m`, `v`, `val`, as suggested previously.)

```suggestion
	m map[IRegisterIndex]Digests
```

comid/integrityregisters_test.go
```diff
@@ -0,0 +1,213 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/veraison/swid"
+)
+
+func prepareRegister(t *testing.T, iType string) (*IntegrityRegisters, error) {
```
thomas-fossati, line 15: it doesn't look like the contents of the `IntegrityRegisters` produced by this function are used anywhere.  Looking at the tests below I suspect if you just returned an empty thing, e.g.:
```golang
	&IntegrityRegisters{
		m: map[IRegisterIndex]Digests{},
	},
```
all tests would still pass :)

comid/integrityregisters_test.go
```diff
@@ -0,0 +1,213 @@
...
+	err := register.AddDigests(uint(0), []swid.HashEntry{})
+	assert.EqualError(t, err, expectedErr)
+}
+
+func TestIntegrityRegisters_MarshalCBOR_UIntIndex_OK(t *testing.T) {
+	reg, err := prepareRegister(t, "uint")
+	require.NoError(t, err)
+	bytes, err := reg.MarshalCBOR()
+	require.NoError(t, err)
+	fmt.Printf("CBOR Payload = %x", bytes)
```
thomas-fossati, line 72: rather than printing to stdout, the test should check that the expected value is produced

comid/integrityregisters_test.go
```diff
@@ -0,0 +1,213 @@
...
+		require.NoError(t, err)
+	}
+}
+
+func TestIntegrityRegisters_MarshalJSON_UIntIndex_OK(t *testing.T) {
+	reg, err := prepareRegister(t, "uint")
+	require.NoError(t, err)
+	bytes, err := reg.MarshalJSON()
+	require.NoError(t, err)
+	fmt.Printf("JSON Payload = %s", bytes)
```
thomas-fossati, line 94: ditto

comid/integrityregisters_test.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func TestIntegrityRegisters_MarshalCBOR_TextIndex_OK(t *testing.T) {
+	reg, err := prepareRegister(t, "text")
+	require.NoError(t, err)
+	bytes, err := reg.MarshalCBOR()
+	require.NoError(t, err)
+	fmt.Printf("CBOR Payload = %x", bytes)
```
thomas-fossati, line 102: ditto

comid/integrityregisters_test.go
```diff
@@ -0,0 +1,213 @@
...
+
```
thomas-fossati, line 103: spurious NL

comid/integrityregisters_test.go
```diff
@@ -0,0 +1,213 @@
...
+		require.NoError(t, err)
+	}
+}
+
+func TestIntegrityRegisters_MarshalJSON_TextIndex_OK(t *testing.T) {
+	reg, err := prepareRegister(t, "text")
+	require.NoError(t, err)
+	bytes, err := reg.MarshalJSON()
+	require.NoError(t, err)
+	fmt.Printf("JSON Payload = %s", bytes)
```
thomas-fossati, line 124: ditto

comid/integrityregisters_test.go
```diff
@@ -0,0 +1,213 @@
...
+		"2":{"key_type":"uint","Value":["sha-256;5Fty9cDAtXLbTY06t+l/No/3TmI0eoJN7LZ6hOUiTXU="]},
+		"3":{"key_type":"text","Value":["sha-256;5Fty9cDAtXLbTY06t+l/No/3TmI0eoJN7LZ6hOUiTXU="]},
+		"4":{"key_type":"text","Value":["sha-256;5Fty9cDAtXLbTY06t+l/No/3TmI0eoJN7LZ6hOUiTXU="]}
+		}`
+	reg := NewIntegrityRegisters()
+	err := reg.UnmarshalJSON([]byte(j))
+	require.NoError(t, err)
+	for index, digests := range reg.Register {
+		_, uok := index.(uint)
+		_, sok := index.(string)
```
thomas-fossati, line 173: this check is weak

comid/integrityregisters_test.go
yogeshbdeshpande: @thomas-fossati : The order in the `map` is not always the same, hence one cannot ascertain `fixed` sequence to compare against to. If we want fixed ordering, I will have to change the implementation to encode it in a fixed manner!

Let me know, what are your views on this, shall we change internal implementation, though in the CoRIM specification we stated that the Register Order is not guaranteed when comparing..

comid/integrityregisters_test.go
yogeshbdeshpande: See above!

comid/integrityregisters_test.go
thomas-fossati: > @thomas-fossati : The order in the `map` is not always the same, hence one cannot ascertain `fixed` sequence to compare against to. If we want fixed ordering, I will have to change the implementation to encode it in a fixed manner!

`cbor.SortCoreDeterministic` should do the trick

> Let me know, what are your views on this, shall we change internal implementation, though in the CoRIM specification we stated that the Register Order is not guaranteed when comparing..

Sure, and the reason for that is the spec should not constrain the implementations.  So that implementations can choose whether to do deterministic encoding or not depending on their tradeoffs.

comid/integrityregisters.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	for _, digest := range digests {
+		if err := i.AddDigest(index, digest); err != nil {
+			return fmt.Errorf("unable to add Digest: %w", err)
+		}
+	}
+	return nil
+}
+
+// AddDigest allows inserting a Digest at a specific registerIndex
```
thomas-fossati, line 40: use the actual parameter names in the docstring

```suggestion
// AddDigest allows inserting a digest at a specific index
```

The docstring should also say what kinds of index types are supported.

comid/integrityregisters_test.go
yogeshbdeshpande: Sure, I will check. How about JSON ? I will check, but I hope there should be a similar determinism for JSON Encoding?

thomas-fossati: > How about JSON ? I will check, but I hope there should be a similar determinism for JSON Encoding?

That is not needed because [assert.JSONEq](https://github.com/stretchr/testify/blob/v1.8.4/assert/assertions.go#L1710) does not care about ordering of map entries.

comid/integrityregisters_test.go
yogeshbdeshpande: Addressed by strict checking!

comid/integrityregisters.go
yogeshbdeshpande: Agree and changes made

comid/integrityregisters.go
yogeshbdeshpande: Made it `concise` but retained explicit setting to be clear

comid/integrityregisters_test.go
yogeshbdeshpande: Use the returned types now!

comid/integrityregisters_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+func TestIntegrityRegisters_MarshalCBOR_TextIndex_OK(t *testing.T) {
+	reg, err := prepareRegister(t, "text")
+	// A5                                   # map(5)
+	// 61                                   # text(1)
+	//    30                                # "0"
+	// 81                                   # array(1)
+	//    82                                # array(2)
+	// 	      01                             # unsigned(1)
+	//  	  58 20                          # bytes(32)
```
thomas-fossati, line 145: we can remove this I think

comid/integrityregisters_test.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+func TestIntegrityRegisters_MarshalCBOR_UIntIndex_OK(t *testing.T) {
+	reg, err := prepareRegister(t, "uint")
+	// A5                                      # map(5)
+	//    00                                   # unsigned(0)
+	//    81                                   # array(1)
+	//       82                                # array(2)
+	//          01                             # unsigned(1)
+	//          58 20                          # bytes(32)
```
thomas-fossati, line 72: this can be removed

thomas-fossati (commented):
couple of spurious blocks to remove

the rest looks good, thanks!

comid/integrityregisters_test.go
yogeshbdeshpande: I think, we can leave it to highlight that line: 141 is how the index is exercised as `strings` within map and how it is different at line 68 where it is exercised as `uint`

yogeshbdeshpande: WDYT?

comid/integrityregisters_test.go
yogeshbdeshpande: Resolved after adding clarification as to what this reflect and its significance!

comid/integrityregisters_test.go
```diff
@@ -0,0 +1,302 @@
...
+func TestIntegrityRegisters_AddDigests_NOK(t *testing.T) {
+	expectedErr := `no digests to add`
+	register := IntegrityRegisters{}
+	err := register.AddDigests(uint(0), []swid.HashEntry{})
+	assert.EqualError(t, err, expectedErr)
+}
+
+func TestIntegrityRegisters_MarshalCBOR_UIntIndex_OK(t *testing.T) {
+	reg, err := prepareRegister(t, "uint")
+	// Below is the partial CBOR Pretty notation to highlight index as uint: unsigned(0) (at line 70)
```
thomas-fossati, line 67: just noting that the "at line 70" comment is going to become stale quite quickly ;-)

PR #109, opened 2024-02-21
Realm Reference Values

yogeshbdeshpande:


cocli/data/corim/templates/corim-cca-realm.json
```diff
@@ -0,0 +1,25 @@
+{
+    "corim-id": "5c57e8f4-46cd-421b-91c9-08cf93e13cfc",
+    "dependent-rims": [
+        {
+            "href": "https://parent.example/rims/ccb3aa85-61b4-40f1-848e-02ad6e8a254b",
+            "thumbprint": "sha-256:5Fty9cDAtXLbTY06t+l/No/3TmI0eoJN7LZ6hOUiTXU="
+        }
+    ],
+    "profiles": [
+        "http://arm.com/cca/realm/1"
```
thomas-fossati, line 10: FYI.  There was a discussion in the PSA Token draft regarding URIs used for naming profiles which deprecated the `http` scheme in favour of the `tag` scheme, to avoid accidental dereferencing.  I think the same logics would apply here.

cocli/data/corim/templates/corim-cca-realm.json
```diff
@@ -0,0 +1,25 @@
+{
+    "corim-id": "5c57e8f4-46cd-421b-91c9-08cf93e13cfc",
+    "dependent-rims": [
+        {
+            "href": "https://parent.example/rims/ccb3aa85-61b4-40f1-848e-02ad6e8a254b",
+            "thumbprint": "sha-256:5Fty9cDAtXLbTY06t+l/No/3TmI0eoJN7LZ6hOUiTXU="
+        }
+    ],
```
thomas-fossati, line 8: This CoRIM feature whose functioning is not especially clear to me ;-) I wouldn't want to trick people into thinking that it's a necessary feature for CCA endorsements.

cocli/data/comid/templates/comid-cca-realm-refval.json
```diff
@@ -0,0 +1,60 @@
...
+                                    ]
+                                },
+                                "rem": {
+                                    "key-type": "text",
+                                    "value": [
+                                        "sha-384;IQe752H8pS2VE2oTVNt6TdV7Gya+DT2nHZ6yOYazS6YVq/ZRTPNeWp6lWgMtBop4",
+                                        "sha-384;JQe752H8pS2VE2oTVNt6TdV7Gya+DT2nHZ6yOYazS6YVq/ZRTPNeWp6lWgMtBop4",
+                                        "sha-384;MQe752H8pS2VE2oTVNt6TdV7Gya+DT2nHZ6yOYazS6YVq/ZRTPNeWp6lWgMtBop4",
+                                        "sha-384;NQe752H8pS2VE2oTVNt6TdV7Gya+DT2nHZ6yOYazS6YVq/ZRTPNeWp6lWgMtBop4"
+                                    ]
```
thomas-fossati, line 51: I don't think this is the correct encoding.  The semantics associated with the multi-value digest is: _"same measured object using different hash algorithm"_.  Here we have multiple separate objects (i.e., REM[0], ..., REM[3]) that are measured using the same algorithm.  We should instead establish a convention to explicitly name the REMs.

thomas-fossati: Maybe that's not entirely clear from the current prose, but that's the intended semantics:

_"Any digests associated with a measured object represent an acceptable state for the object."_

so, the encoding used above means: the measured object "rem" can have m[1] or m[2] or m[2] or m[3], which is not what we want to say.

There are four (or less, or more) distinct REM objects, each with its own set of acceptable values.

yogeshbdeshpande: If I understand correctly for `REM` can have 1-4 Measurements in any order is the flexibility provided by the system.

There is no REM-0 or REM-1, like that: It is just a collection with a 1-4 entries. 

I interpreted it based on the RMM spec hence I encoded it like this!

thomas-fossati: no, REM banks have (implicit) names, i.e., their index.

It's not a collection, it's an ordered list.

When you do call the `RSI_MEASUREMENT_EXTEND` ABI you need to explicitly pass the index (that is, the name) of the bank which you want to extend.

thomas-fossati: In particular, we need to be able to express semantics like: _"I only care that bank 2 is `0x3996003d486fb91ffb056f7d03f2b2992b215b31dbe7af4b373431fc7d319da3`"_ (if using SHA-256), or `11fb303c8d002309999c2ac9b6897bc294de3b76ab7664f024ec4b37552f0508b34f7390114bf09f93f73d508c57b0ef79d6b2fabbc0f3ad87f3c1ccef7d9be4` (if using SHA-512).

cocli/data/corim/templates/corim-cca-realm.json
yogeshbdeshpande: Thanks,  I follow that:
However, there should be some degree of consistency in our approach.
Otherwise the profile definition should look very much fragmented.

Either we change the same for CCA Token Profiles and CCA Platform Endorsement Profile at the same time 
and do it in one batch otherwise it is super confusing to everyone!

cocli/data/comid/templates/comid-cca-realm-refval.json
yogeshbdeshpande: I think, it is treated as a `Set` Not Individual entries, like REM-0. 
Also, as per our philosophy, Reference Values should be aligned as much close to how they appear in the Evidence.

cocli/data/corim/templates/corim-cca-realm.json
yogeshbdeshpande: Created Issue #110 to address this in future changes

thomas-fossati (approved):
Thanks for addressing my comments.

PR #111, opened 2024-04-16
Provide missing interface implementations

yogeshbdeshpande:
As part of adding extension enhancement to `CoRIM`, part of the interfaces for CoMID library were omitted.
These are required and is been invoked from `veraison/services`.


setrofim (approved):
LGTM!

PR #117, opened 2024-07-03
fix: use value receivers for MarshalXXX methods

setrofim:
Use value, rather than pointer, receivers for MarshalCBOR and MarshalJSON methods. Since those methods don't mutate the structures, pointers are not necessary, and using them prevents the methods from being invoked when marshaling an instance of that structure (rather than a pointer to one). This is not a huge deal as CoRIM fields typically use pointers, however it can lead to surprising results when playing with the structures on their own.

### DIFF
--- a/comid/classid.go
+++ b/comid/classid.go
@@ -293,8 +293,8 @@ func (o TaggedImplID) Bytes() []byte {
 	return o[:]
 }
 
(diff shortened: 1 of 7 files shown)


thomas-fossati (approved):
LGTM

Issue #116, opened 2024-06-06
CoMID generates incorrect triple index for `identity-triples` and `attest-key-triples`


amd-isaac:
When generating a CoMID with an entry in `identity-triples` (AKA `dev-identity-keys` in the `cocli` CoMID template), the resultant binary uses a `triples-map` index of `3` rather than the value of `2` in the [IETF draft spec](https://www.ietf.org/archive/id/draft-ietf-rats-corim-04.html#section-3.1.4-4.3.1). Similarly the `attest-key-triples` generates with an index of `2` when the spec says it should be `3`. Perhaps these two were swapped during the initial implementation? Or is the change in terminology intentional?

Code from `triples.go`: 
```
type Triples struct {
    ReferenceValues *[]ReferenceValue `cbor:"0,keyasint,omitempty" json:"reference-values,omitempty"`
    EndorsedValues  *[]EndorsedValue  `cbor:"1,keyasint,omitempty" json:"endorsed-values,omitempty"`
    AttestVerifKeys *[]AttestVerifKey `cbor:"2,keyasint,omitempty" json:"attester-verification-keys,omitempty"`
    DevIdentityKeys *[]DevIdentityKey `cbor:"3,keyasint,omitempty" json:"dev-identity-keys,omitempty"`

    Extensions
}
```

Minimal example `template.json` file:
```
{
  "tag-identity": {
    "id": "eedf041b-2dbe-4f1a-968f-cc64f21db8ae"
  },
  "entities": [
    {
      "name": "Tag Creator Jr.",
      "roles": [
        "tagCreator"
      ]
    }
  ],
  "triples": {
    "dev-identity-keys": [
      {
        "environment": {
          "class": {
            "vendor": "Vendor Esq."
          }
        },
        "verification-keys": [
          {
            "type": "pkix-base64-key",
            "value": "-----BEGIN PUBLIC KEY-----\nMIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQD2mhCBSTYrACSHdLtbl3McOoPm\nZYQX40XjPbzKFsnc/5lKxjjVb3EAVbOm9MXOI+0PS/6iKhwTKox4d7LoY4qHvIZ1\nZka2tYxUxRqXLzzcWlUvopuH/O5S7kgQc4IU7+3KZ1dnacsHcEzBpO4uhapjCn7z\nUdWI3g/M4Xz8EQGRlQIDAQAB\n-----END PUBLIC KEY-----"
          }
        ]
[...]

PR #118, opened 2024-07-15
Fix map extensions and implement profiles

setrofim:
This pull request redesigns the handling of CoRIM map extensions to make them actually usable by allowing them to be registered with, and propagated down from, higher-level structures. To ensure correct handling of extensions when marshalling slices, they are wrapped in a custom container type that manages extensions on behalf of its values.

Profiles can be used to associate a set of map extensions with an `eat.Profile`, and have them automatically be registered on (un)marshalling.

A number of smaller fixes and refactors are also implemented (see individual commit messages).


setrofim:
This addresses https://github.com/veraison/corim/issues/116

cots/example_test.go
```diff
@@ -1,3 +1,5 @@
+// Copyright 2024 Contributors to the Veraison project.
```
yogeshbdeshpande, line 1: Are these old files with missing Copyright Headers ?

If yes, then we should introduce the 2023-2024, as these files were introduced by Sabreen in 2023

cots/hardware_version_type.go
```diff
@@ -1,3 +1,5 @@
+// Copyright 2024 Contributors to the Veraison project.
```
yogeshbdeshpande, line 1: same comment as above!

encoding/embedded.go
```diff
@@ -1,3 +1,5 @@
+// Copyright 2024 Contributors to the Veraison project.
```
yogeshbdeshpande, line 1: same here!

yogeshbdeshpande:
@setrofim :  I assume that you would not be squashing commits and should keep them separate after review as well.

The reason been quite a bit of un-related fixes have gone in (for example: swap DevIdentity Keys and AttestVerif Keys), which is un-related to Extensions work which was the intent of the PR ?

yogeshbdeshpande:
Also request please spell check and review, all commit messages, I noticed few typos in them, which un-fortunately I cannot comment.

setrofim:
> @setrofim : I assume that you would not be squashing commits and should keep them separate after review as well.
> 
> The reason been quite a bit of un-related fixes have gone in (for example: swap DevIdentity Keys and AttestVerif Keys), which is un-related to Extensions work which was the intent of the PR ?

Correct. I never rely of GitHub commit sqashing. Any incidental fix-ups have already been "pre-squashed", add the same will go for any further changes due to review feedback.

Each commit represents a single logical change (or a set of tightly-coupled changes) detailed in its commit message, and they should be rebased on top of `main` as-is.

setrofim:
> If yes, then we should introduce the 2023-2024, as these files were introduced by Sabreen in 2023

done...

>Also request please spell check and review, all commit messages, I noticed few typos in them, which un-fortunately I cannot comment.

...and done (I believe).

yogeshbdeshpande:
- Commit Number : 
2992d5a LGTM!

comid/comid.go
```diff
@@ -214,10 +214,10 @@ func (o *Comid) AddEndorsedValue(val ValueTriple) *Comid {
 
 // AddAttestVerifKey adds the supplied endorsed value to the
 // attest-key-triples list of the target Comid.
-func (o *Comid) AddAttestVerifKey(val AttestVerifKey) *Comid {
+func (o *Comid) AddAttestVerifKey(val KeyTriple) *Comid {
```
yogeshbdeshpande, line 217: While you are at 217 may be please correct the comment on Line 215, it is perhaps a `Verification key` rather than an Incorrect `endorsed value` due to a copy paste error

comid/keytriple.go
```diff
@@ -1,21 +1,21 @@
-// Copyright 2021-2023 Contributors to the Veraison project.
+// Copyright 2021-2024 Contributors to the Veraison project.
 // SPDX-License-Identifier: Apache-2.0
 
 package comid
 
 import "fmt"
 
-// DevIdentityKey stores an identity-triple-record with CBOR and JSON
+// KeyTriple stores an identity-triple-record with CBOR and JSON
```
yogeshbdeshpande, line 8: Is line 8 a fair statement, now that KeyTriple is a common type for both Identity and Attestation Key ?

setrofim: Good point. Updated to "KeyTriple stores a cryptographic key triple record (identity-triple-record
or attest-key-triple-record) with CBOR and JSON"

corim/testcases/src/corim-with-extensions.yaml
```diff
@@ -0,0 +1,75 @@
+# minimalist unsigned-corim that embeds comid.PSARefValJSONTemplate 
+# - profile (claim 3) has been set to "http://example.com/test-profile"
+# - extension claim -1 has been added to corim
+# - extension claim -1 has been added to comid entity
```
yogeshbdeshpande, line 4: It is not very clear how these extensions have been realised, can you please add a comment next to the extension. 

If I understand correctly, Line 3 comment applies to Line 8: where foo has been added?

However, for Entity, inside CoMID, -1 is a number, is it Reg-ID or something ?

May be work adding the YAML comment : Please..

corim/unsignedcorim_test.go
```diff
@@ -72,9 +67,9 @@ func TestUnsignedCorim_AddComid_and_marshal(t *testing.T) {
 
 	fmt.Printf("CBOR: %x", actual)
 
-	expected := testGoodUnsignedCorim
+	expected := testGoodUnsignedCorimCBOR
```
yogeshbdeshpande, line 70: For readers sake, now please refer to link, where these variables are taken from ?

setrofim: ? we don't do that for any other commonly defined test variables, and this is something that is aready easily resolved by the IDE.

corim/testcases/src/corim-with-extensions.yaml
setrofim: Well, the whole point is that they do not have pre-defined names, in the spec so -1 is just -1;

setrofim: (they do not have pre-defined numbers either, which is why -1, the first in the extension range, is used for both corim and comid entity.

yogeshbdeshpande: All I am saying is to state how you are using -1 in the code, as an example!. I am not sure it is very clear how extensions are exercised...

setrofim: This is not indened as an example of extension usage. This is a test case for a very specific test. The test just checks that the  code can access the extensions embedded in the struct heiarchy. It doesn't do anything with the extesions, and so there is no meaning attached to the extensions beyound "this is the extension with key -1 and value "foo"". 

 Bear in mind the the YAML files are intended purely as an easier means to getting at binary blobs for tests (and binary blobs is all we had before -- those are even less clear). We do not expect people using the extensions mechanim to be using YAML (they will be using the `cocli` "templates" or defining things programatically in code). There is a separate file (`corim/example_profile_test.go`) that actually contains example code intended to show how the extensions are defined and used.

comid/comid.go
```diff
@@ -252,26 +272,50 @@ func (o Comid) ToCBOR() ([]byte, error) {
 		return nil, err
 	}
 
-	return encoding.SerializeStructToCBOR(em, &o)
+	var entities *Entities
+	if o.Entities != nil && o.Entities.IsEmpty() {
```
yogeshbdeshpande, line 276: The whole block is extremely confusing here. Suggest renaming to some better naming to clarify the concept!

yogeshbdeshpande: Also I am not clear as to why this is not an error? Essentially Entire Collection object is empty however a pointer is set ?

comid/comid.go
```diff
@@ -252,26 +272,50 @@ func (o Comid) ToCBOR() ([]byte, error) {
...
-
 // ToJSON serializes the target Comid to JSON
 func (o Comid) ToJSON() ([]byte, error) {
 	if err := o.Valid(); err != nil {
 		return nil, err
 	}
 
-	return encoding.SerializeStructToJSON(&o)
+	var entities *Entities
+	if o.Entities != nil && o.Entities.IsEmpty() {
```
yogeshbdeshpande, line 302: Same here!

extensions/collection.go
```diff
@@ -0,0 +1,225 @@
...
+// Map is a mapping of extension Point's to IMapValue's (pointers to structs that
+// define extension fields.
+type Map map[Point]IMapValue
+
+// NewMap instantiates an empty Map
+func NewMap() Map {
+	return make(Map)
+}
+
+// Add a Paint-IMapValue entry to the map. If there is already an IMapValue
```
yogeshbdeshpande, line 30: Typo `Paint` to Point

extensions/collection.go
```diff
@@ -0,0 +1,225 @@
...
+// Collection is a generic container for objects who's pointers implement
+// IExtensible. In addition of containing instances of extensible objects, it
+// is capable of registering extensions for those instances.
+type Collection[P any, I IExtensible[P]] struct {
+	Values []P
+
+	valueExtensions cache
+}
+
+// NewCollection instantiates an empty Measurements array
```
yogeshbdeshpande, line 56: Is the comment on line, correct ? I thought it is generic to any extensible array?

extensions/collection.go
```diff
@@ -0,0 +1,225 @@
...
+func NewCollection[P any, I IExtensible[P]]() *Collection[P, I] {
+	return &Collection[P, I]{Values: []P{}, valueExtensions: cache{make(Map)}}
+}
+
+// Add the supplied Measurement to the target Measurement
```
yogeshbdeshpande, line 61: Same here!

extensions/collection.go
```diff
@@ -0,0 +1,225 @@
...
+	o.Values = []P{}
+}
+
+// GetExtensions returns the extensions IMapValue that has been registered with
+// the collection.
+func (o *Collection[P, I]) GetExtensions() IMapValue {
+	return o.valueExtensions.Get()
+}
+
+// Valid returns an error if the ReferenceValues collection is invalid, i.e. if it
```
yogeshbdeshpande, line 100: Similarly on Line 100, does not reflect generality?

yogeshbdeshpande (commented):
Code Review is WIP

comid/comid.go
setrofim: `Entities` is a `Collection`, as well as containing elements of some type, it also caches extensions for those elements. If extensions have been registered, but no elements have been added, the collection will be empty but it will still exist to keep track of the extensions (the pointer won't be `nil`). 

Since we want to omit the claim, rather than marshaling an empty list, we temporarily set the field to `nil`, let the normal marhaling take place, and then restore it afterwards.

I can rename to local vairable to something like `temp` or `save` but I'm not really sure that would make things clearer?

extensions/collection.go
setrofim: Argh, yeah, I started out doing this for measurements, and then realised that I'd need to do that for every extensible type, so made it generic. the comment is left over from old implmentation; will fix.

comid/comid.go
setrofim: Oh! actually, since teh receiver is value and not pointer, we don't actually need to save a local copy. We can just set it to nil. Since the `o` that exists in this method was passed by value and is a copy of the caller, modifying the former won't affect the latter.

comid/comid.go
```diff
@@ -252,6 +272,16 @@ func (o Comid) ToCBOR() ([]byte, error) {
 		return nil, err
 	}
 
+	// If extensions have been registered, the collection will exist, but
+	// might be empty. If that is the case, set it to nil to avoid
+	// marshaling an empty list (and let he marshaller omit the claim
```
yogeshbdeshpande, line 277: `(and let the`

comid/entity_test.go
```diff
@@ -50,24 +50,24 @@ func TestEntity_Valid_name_regid_and_roles(t *testing.T) {
 }
 
 func TestEntities_Valid_empty(t *testing.T) {
-	e := Entity{}
-	tv := NewEntities().AddEntity(e)
+	e := &Entity{}
```
yogeshbdeshpande, line 53: Since the Extensions are now built in the collection, i.e. Entities as you say, should we also not add the Unit Tests to exercise the examples for extensions in `entity_test.go` Far easier for readers to understand at one place, 
- How Native Entity is tested
- How Extension in the Entity is tested ?

comid/measurement.go
```diff
@@ -375,6 +390,16 @@ func (o *Mval) UnmarshalCBOR(data []byte) error {
 
 // MarshalCBOR serializes to CBOR
 func (o Mval) MarshalCBOR() ([]byte, error) {
+	// If extensions have been registered, the collection will exist, but
+	// might be empty. If that is the case, set it to nil to avoid
+	// marshaling an empty list (and let he marshaller omit the claim
```
yogeshbdeshpande, line 395: `(and let the`

comid/measurement.go
```diff
@@ -385,6 +410,16 @@ func (o *Mval) UnmarshalJSON(data []byte) error {
 
 // MarshalJSON serializes to JSON
 func (o Mval) MarshalJSON() ([]byte, error) {
+	// If extensions have been registered, the collection will exist, but
+	// might be empty. If that is the case, set it to nil to avoid
+	// marshaling an empty list (and let he marshaller omit the claim
```
yogeshbdeshpande, line 415: `(and let the`

comid/entity_test.go
setrofim: `Entities` is a thin wrapper around `Collection`; there is nothing to test -- it just calls through to `Collection` for everything. Collection is tested by `collection_test.go`.

For example of extensions we have the `example_profile_test.go`.

yogeshbdeshpande: May be I was not clear, let me try to explain it again:
What I meant was multiple extensions via Entities, i.e. `Instantiate` Two Entity Structures via Collections, each have a different extension inside it, and set, get and assert the extensions are exercised correctly, inside Entities Collection. This way, readers of Entity know: how to exercise Entity specific Extensions.

Also then there is also a case for not using Extensions but still how to set Multiple Entities, - with no extensions. 

Perhaps this should have been covered outside of the scope of this PR, but might be helpful to add it at some point in time?

comid/measurement.go
```diff
@@ -728,10 +771,50 @@ func (o Measurement) Valid() error {
...
 
-	// check non-empty<> condition on measurement-values-map
-	if err := o.Val.Valid(); err != nil {
-		return err
-	}
+	return o.Val.Valid()
+}
 
-	return nil
+// Measurements is a container for Measurement instances and their extensions.
```
yogeshbdeshpande, line 777: The whole Measurements was `in Measurements.go` file, any reason why this is moved here ?

The file already is very long...?

comid/referencevalue.go
```diff
@@ -1,24 +1,86 @@
-// Copyright 2021 Contributors to the Veraison project.
+// Copyright 2021-2024 Contributors to the Veraison project.
 // SPDX-License-Identifier: Apache-2.0
 
 package comid
 
-import "fmt"
+import (
```
yogeshbdeshpande, line 6: The file name no longer match the contents of the file, earlier it was correctly done.

Now the name should be valuetriples.go

comid/referencevalue.go
```diff
@@ -1,24 +1,86 @@
...
+	"errors"
+	"fmt"
 
-type ReferenceValue struct {
+	"github.com/veraison/corim/extensions"
+)
+
```
yogeshbdeshpande, line 12: Now that we have a common triples semantics, best to add a comment to state that 

// ValueTriple is a common structure that carries the information for Reference Values and Endorsed Values as of V4 of the specification

comid/measurement.go
setrofim: It's under 1000 lines, that's hardly "long" :). 

The implementation of `Measurements` is trivial, and it doesn't seem to make much sense spliting that out when the rest of measurement-related stuff is in here. Having things spread a cross a multitude of files, each containing one struct just makes things more annoying to manage.

yogeshbdeshpande:
@setrofim : A General suggestion: We should add a test case, which adds `multiple` custom extensions i.e say one for Measurements (MVal) and one for say Triples, demonstrate via example, how it is (a) Set (b) Get (c) CBOR and JSON Encoded.
It is fine to do it in this PR or a separate PR, but will be certainly helpful to new-joiners!

comid/referencevalue.go
yogeshbdeshpande: Thank you!

comid/entity_test.go
setrofim: This scenario is already covered by `exampole_profile_test.go` (with mesaurements, rather than entities, but the same principle applies).

setrofim:
> @setrofim : A General suggestion: We should add a test case, which adds `multiple` custom extensions i.e say one for Measurements (MVal) and one for say Triples, demonstrate via example, how it is (a) Set (b) Get (c) CBOR and JSON Encoded/Decoded It is fine to do it in this PR or a separate PR, but will be certainly helpful to new-joiners!

This is covered by `example_profile_test.go` (note that examples are also executed as part of the test suite, so they  double as tests).

corim/example_profile_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+// ----- profile definition -----
+// The following code defines a profile with the following extensions and
+// constraints:
+// - Entities may contain an address field
+// - Reference values may contain a Unix timestamp indicating when the
+//   individual measurement was taken.
+// - The language claim (CoMID index 0) must be present, and must be "en-GB"
+
+// These extensions will be used for both CoMID and CoRIM entities.
```
yogeshbdeshpande, line 27: Minor comment: Change // These extensions to 
// EntityExtensions will be used for both..

corim/example_profile_test.go
```diff
@@ -0,0 +1,199 @@
...
+		extractedCorim.Tags[0],
+		extractedCorim.Profile,
+	)
+	if err != nil {
+		log.Fatalf("could not unmarshal corim: %v", err)
+	}
+
+	fmt.Printf("Language: %s\n", *extractedComid.Language)
+	fmt.Printf("Entity: %s\n", *extractedComid.Entities.Values[0].Name)
+	fmt.Printf("        %s\n", extractedComid.Entities.Values[0].
```
yogeshbdeshpande, line 103:
```suggestion
	fmt.Printf(" Address:       %s\n", extractedComid.Entities.Values[0].
```

corim/example_profile_test.go
```diff
@@ -0,0 +1,199 @@
...
+	// Language: en-GB
+	// Entity: ACME Ltd.
+	//         123 Fake Street
+	// Measurements:
+	//     87428fc522803d31065e7bce3cf03fe475096631e5e07bbd7a0fde60c4cf25c7 taken at 2024-07-12T11:03:10
+	//     0263829989b6fd954f72baaf2fc64bc2e2f01d692d4de72986ea808f6e99813f taken at 2024-07-12T11:03:10
+	//     a3a5e715f0cc574a73c3f9bebb6bc24f32ffd5b67b387244c2c909da779a1478 taken at 2024-07-12T11:03:10
+}
+
+// note: this example is rather verbose as we're going to be constructing a
```
yogeshbdeshpande, line 126:
```suggestion
// note: this example is rather verbose as we're going to construct a
```

corim/example_profile_test.go
```diff
@@ -0,0 +1,199 @@
...
+// CoMID by hand. In practice, you would typically write a JSON document and
+// then unmarshal that into a CoRIM before marshaling it into CBOR (in which
```
yogeshbdeshpande, line 128: Did you mean `CoMID` on line 128

corim/example_profile_test.go
```diff
@@ -0,0 +1,199 @@
...
+		comid.MustCreatePSARefValID(
+			comid.TestSignerID, "BL", "5.0.5",
+		)).AddDigest(swid.Sha256_32, []byte{0xab, 0xcd, 0xef, 0x00})
+
+	// alternatively, we can add extensions to individual value before
+	// adding it to the collection. Note that because we're adding the
+	// extension directly to the measurement, we're using a different
+	// extension point, comid.ExtMval rather than comid.ExtReferenceValue,
+	// as a measurement doesn't know that its going to be part of reference
+	// value, ans so is unaware of reference value extension points.
```
yogeshbdeshpande, line 174:
```suggestion
	// value, and so is unaware of reference value extension points.
```

corim/profiles.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// AllExtensionPoints is a list of all valid extension.Point's
+var AllExtensionPoints = make(map[extensions.Point]bool) // populated inside init() below
+
+// UnmarshalSignedCorimFromCBOR unmarshals a SignedCorim from provided
+// CBOR data. If there are extensions associated with the profile specified by
+// the data, they will be registered with the UnsignedCorim before it is
+// unmarshaled.
+func UnmarshalSignedCorimFromCBOR(buf []byte) (*SignedCorim, error) {
```
yogeshbdeshpande, line 49: The names sounds duplicate to the similar functions given in the unsignedcorim.go `FromCBOR()` Method. Perhaps, we should clarify and also if possible re-name them to remove the confusion.

corim/profiles.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	return ret, nil
+}
+
+// UnmarshalUnsignedCorimFromCBOR unmarshals an UnsignedCorim from provided
+// CBOR data. If there are extensions associated with the profile specified by
+// the data, they will be registered with the UnsignedCorim before it is
+// unmarshaled.
+func UnmarshalUnsignedCorimFromCBOR(buf []byte) (*UnsignedCorim, error) {
```
yogeshbdeshpande, line 76: Same comment as in Line 49

extensions/README.md
```diff
@@ -47,22 +47,48 @@ types. This is done via three distinct extension mechanisms:
...
-
-- `comid.Comid`
-- `comid.Entity`
-- `comid.FlagsMap`
-- `comid.Mval`
-- `comid.Triples`
-- `corim.Entity`
-- `corim.Signer`
-- `corim.UnsignedCorim`
+can be identified by the embedded `Extensions` struct. Each extensible type has
```
yogeshbdeshpande, line 50: Just a quick question: Line 2 of the README.md refer to V2 of the specification, are you adhereing to V4 or still on V2?

extensions/README.md
```diff
@@ -47,22 +47,48 @@ types. This is done via three distinct extension mechanisms:
...
+| `comid.Comid`         | `comid.ExtComid`                                              |
+| `comid.Entity`        | `comid.ExtEntity`                                             |
+| `comid.FlagsMap`      | `comid.ExtReferenceValueFlags`, `comid.ExtEndorsedValueFlags` |
+| `comid.Mval`          | `comid.ExtReferenceValue`, `comid.ExtEndorsedValue`           |
+| `comid.Triples`       | `comid.ExtTriples`                                            |
+| `corim.Entity`        | `corim.ExtEntity`                                             |
+| `corim.Signer`        | `corim.ExtSigner`                                             |
+| `corim.UnsignedCorim` | `corim.ExtUnsignedCorim`                                      |
+
+Note that `comid.Mval` and `comid.FlagsMap` are used for both reference and
```
yogeshbdeshpande, line 64:
```suggestion
Note that `comid.Mval` and `comid.FlagsMap` are used for both reference values and
```

extensions/README.md
```diff
@@ -47,22 +47,48 @@ types. This is done via three distinct extension mechanisms:
 
 Map extensions allow extending CoRIM maps with additional keys, effectively
```
yogeshbdeshpande, line 48: Map extensions allow extending CoRIM and CoMID maps with additonal keys,

extensions/README.md
```diff
@@ -47,22 +47,48 @@ types. This is done via three distinct extension mechanisms:
...
+| `comid.Mval`          | `comid.ExtReferenceValue`, `comid.ExtEndorsedValue`           |
+| `comid.Triples`       | `comid.ExtTriples`                                            |
+| `corim.Entity`        | `corim.ExtEntity`                                             |
+| `corim.Signer`        | `corim.ExtSigner`                                             |
+| `corim.UnsignedCorim` | `corim.ExtUnsignedCorim`                                      |
+
+Note that `comid.Mval` and `comid.FlagsMap` are used for both reference and
+endorsed values, which may be extended separately. This is why there are two
+extension points associated with each. Additionally, `comid.ExtMval` and
+`comid.ExtFlags` also exist for when you want to register extensions with a
```
yogeshbdeshpande, line 67:
```suggestion
`comid.ExtFlags` also exist when you want to register extensions with a
```

extensions/README.md
```diff
@@ -47,22 +47,48 @@ types. This is done via three distinct extension mechanisms:
...
+not meant to accurately reflect the code -- e.g. it omits the container types,
+representing them as slices)
 
 To extend the above types, you need to define a struct containing your
-extensions and pass a pointer to an instance of that struct to the
-`RegisterExtensions()` method of the corresponding instance of the type that is
-being extended. This should be done as early as possible, before any marshaling
-is performed.
+extensions for each extension point that you want to extend. You then need to
+create an `extensions.Map` that maps the extension points  to a pointer to an
```
yogeshbdeshpande, line 83:
```suggestion
create an `extensions.Map` that maps the extension points to a pointer to an
```

extensions/README.md
```diff
@@ -47,22 +47,48 @@ types. This is done via three distinct extension mechanisms:
...
+instance of the corresponding struct. Finally, you need to pass  the map to the
+`RegisterExtensions()` method of an instance of the top-most type that is being
+extended. For example, if you want to extend the comid, reference values, and
+entities, you would create a struct defining the extensions for each, call
+`extensions.NewMap()` to create a new map, call `Add()` on the map to add
+mappings from the three extension points to pointers to empty instances of your
+structs, and, finally, pass the map to `comid.Comid.RegisterExtensions()`. This
+should be done as early as possible, before any marshaling is performed.
```
yogeshbdeshpande, line 91: Should you be referring to the snippet here which one should refer as an example?

extensions/README.md
```diff
@@ -166,16 +194,83 @@ func main() {
...
 	fmt.Printf("also entity email: %s\n", exts.Email)
 }
 ```
 
+### Profiles
+
+Map extensions may be grouped into profiles. A profile is registered,
+associating an `eat.Profile` with an `extensions.Map`. A registered profile can
+be obtained by calling `corim.GetProfile()`, which returns a `corim.Profile`
+object which can can be used to obtain `corim.UnsignedCorim`,
```
yogeshbdeshpande, line 212:
```suggestion
object which can be used to obtain `corim.UnsignedCorim`,
```

yogeshbdeshpande (approved):
Left some Minor .Nits and left one question.

In general LGTM!

extensions/README.md
setrofim: "for when" is the correct construct here as it implies purpose (i.e. they don't just exist at the time you want to extend `Mval` and `Measurement`, but _for_ that time).

extensions/README.md
setrofim: I'll update it to 04, but technically, the current implementation doesn't really fully adhere to any particular rev. E.g. we're only now fixing the profile claim in this pull request, and that was incorrect since rev. 00.

corim/profiles.go
setrofim: The similarity is the point, as they are parallels of each other -- one is a function, the other is a method. You use one when you don't already have an instance to unmarshal into, and they other when you do.

corim/profiles.go
setrofim: ditto as above

corim/example_profile_test.go
setrofim: I did. We're also constructing a `CoRIM` by hand, but that is simple. Most of the verbosity comes from the `CoMID`.

corim/example_profile_test.go
setrofim: What's wrong with the existing phrasing? The imperfect tense seems more appropriate, as the comment is concerned with the process, not the end result (i.e. the point isn't that we're gonna construct a CoMID, but that we will be doing so by hand).

corim/example_profile_test.go
setrofim: The address is part of the Entity, not a sparte entry.

corim/example_profile_test.go
yogeshbdeshpande: That is fine, then you want to say:
this example is rather verabose as we are constructing a

extensions/README.md
yogeshbdeshpande: ok, thanks, this was just a quick check!

corim/example_profile_test.go
setrofim: As the comment before the example, we haven't started yet, so we're going to be doing that, not presently doing that.

Your suggested phrasing would make sesne if the comment was in the middle of the example.

extensions/corim-map-extensions.png
thomas-fossati, line 1: beautiful!  Is it handmade?

corim/testcases/regen-from-src.sh
```diff
@@ -0,0 +1,30 @@
...
+# Copyright 2024 Contributors to the Veraison project.
+# SPDX-License-Identifier: Apache-2.0
+set -e
+
+GEN_TESTCASE=$(go env GOPATH)/bin/gen-testcase
+
+if [[ ! -f ${GEN_TESTCASE} ]]; then
+	echo "installing gen-testcase"
+	go install github.com/veraison/gen-testcase@v0.0.1
+fi
```
thomas-fossati, line 11: Just a curiosity: why didd we need a new tool to synthesise test cases rather than just using diag2cbor.rb as we do, e.g., in https://github.com/veraison/psatoken/tree/main/testvectors/cbor ?

thomas-fossati (approved):
LGTM, thanks!

(I suggest adding Jag as a reviewer.)

setrofim:
> (I suggest adding Jag as a reviewer.)

This has already been shared with Jag in the last public meeting.

PR #122, opened 2024-08-07
fix: propagate ca_cert to OAuth2 authenticator

setrofim:
Use ca_cert certs when configuring the OAuth2 authenticator, in addition to their current use in connection to the API service.

### DIFF
--- a/cocli/cmd/root.go
+++ b/cocli/cmd/root.go
@@ -73,6 +73,7 @@ func initConfig() {
 			"token_url":     v.GetString("token_url"),
 			"username":      v.GetString("username"),
 			"password":      v.GetString("password"),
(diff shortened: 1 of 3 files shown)


cocli/cmd/root.go
```diff
@@ -73,6 +73,7 @@ func initConfig() {
 			"token_url":     v.GetString("token_url"),
 			"username":      v.GetString("username"),
 			"password":      v.GetString("password"),
+			"ca_certs":      v.GetStringSlice("ca_cert"),
```
thomas-fossati, line 76: plural: s/ca_cert/ca_certs/ ?

thomas-fossati (approved):

Issue #123, opened 2024-08-09
ValueTriple has "measurements" should be "measurement" to match corim draft


deeglaze:
The reference-triple-record has 1 environment-map and 1 measurement-map. The ValueTriple's "measurements" collection does not follow that document structure, so I don't know how this produces a valid comid.

It would certainly seem to me that 1 environment could have multiple measurements of different names and value kinds the way this code allows for, but the draft doesn't allow for that. The mkey I thought was a way for multiplicity to be allowed, but also no.

setrofim:
Looks like the code matches [rev00](https://www.ietf.org/archive/id/draft-birkholz-rats-corim-00.html) of the draft, and the triples structure was never brought into alignment with the later versions. We need to fix this.

thomas-fossati:
Noting that this change would impact how PSA and CCA reference values are currently encoded.

setrofim:
> Noting that this change would impact how PSA and CCA reference values are currently encoded.

Yup, this will have ripple effect on a lot of existing code. Fixing this is not going to be trivial...

yogeshbdeshpande:
Yes, this was a known issue however, we had not scoped it fully so had put on hold for a while.

Has impact on PSA, CCA and also Parsec-CCA Scheme under Veraison

setrofim:
@deeglaze this should be resolved by https://github.com/veraison/corim/pull/124

PR #124, opened 2024-08-14
fix(comid): bring ValueTriple into alignment with rev05

setrofim:
At rev00, for reference/endorsed value triples, the CoRIM spec associated an environment with one or more measurements, allowing for two layers of multiplicity.

Since rev01, and up to (current) rev05, the spec removed the second layer, now associating an environment with exactly one measurement.

This commit brings our implementation into alignment with the spec, replacing ValueTriple.Measurements collection with the single ValueTriple.Measurement.

This addresses https://github.com/veraison/corim/issues/123


thomas-fossati (approved):
thanks!

yogeshbdeshpande (approved):
LGTM!

PR #138, opened 2024-12-24
Update README.md with dependency instructions

### LINKED ISSUE #85: Could owner give some requirements  before build ? I have no idea where to start
1. OS
2. tool list need to intall before run this
3. how to build

deeglaze:
A reasonable attempt to close #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ make presubmit
 ```
 and check its output to make sure your code coverage figures are in line with the set target and that there are no newly introduced lint problems.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,8 @@ make presubmit
 ```
 and check its output to make sure your code coverage figures are in line with the set target and that there are no newly introduced lint problems.
 
+Note: You will need mockgen for testing: `go install github.com/golang/mock/mockgen@v1.5.0`, and the usual `go mod tidy` to fetch dependencies.
```
thomas-fossati, line 21: `mockgen` is a standalone CLI: I don't think we need to update the module's dependencies after installing it.

PR #141, opened 2024-12-24
fix (CoRIM Extensions) : Incorrect Assumptions on Measurement Values Map validity

### LINKED ISSUE #134: CoRIM Extensions: Incorrect Assumptions on Measurement Values Map validity
The current code Valid method of MVal struct (measurement.go) assumes that Extensions can only be on top of at least one base type (defined in standards) to be present! Hence it returns with an Error if ONLY Extensions are present with no base types defined.

However, there can be cases (like Intel Profile) where ONLY extensions are present with not even a single base type (from MVal struct) present

Priyanshuthapliyal2005:
Fixes #134


comid/measurement.go
```diff
@@ -436,7 +437,10 @@ func (o Mval) Valid() error {
 		o.UEID == nil &&
 		o.UUID == nil &&
 		o.IntegrityRegisters == nil {
-		return fmt.Errorf("no measurement value set")
+
+		if o.Extensions.IsEmpty() {
```
deeglaze, line 441: Why not include this in the conjunction? I don’t see a reason for the nested if.

Priyanshuthapliyal2005: I uses nested if statements for simpler view of the conditions. However, if using a conjunction is preferred for better readability, I can change it accordingly.

deeglaze: I would prefer it. I don’t think there’s anything about this condition that needs special attention

Priyanshuthapliyal2005: Thanks for the feedback!  it's cleaner with conjunctions. I've updated the code

comid/measurement.go
```diff
@@ -424,6 +424,7 @@ func (o Mval) MarshalJSON() ([]byte, error) {
 }
 
 func (o Mval) Valid() error {
+	 // Check if all base types are nil and Extensions are not nil
```
deeglaze, line 427: Please move this to be a method comment, “Valid returns true if all […]” since Valid is exported but doesn’t have a comment.

deeglaze (approved):

comid/measurement.go
Priyanshuthapliyal2005: I have updated the Valid method to include a comment and moved the check for empty extensions as suggested.

setrofim (approved):

yogeshbdeshpande (approved):
LGTM! Thanks for making the change!

Priyanshuthapliyal2005:
Hi @yogeshbdeshpande  @deeglaze  @setrofim 

Could you please help merge [this PR](https://github.com/veraison/corim/pull/141)? All checks have passed and the PR has been approved by all of you.

Thank you!

deeglaze:
Done

Priyanshuthapliyal2005:
> Done

Thank you so much

Issue #91, opened 2023-09-12
merge the lint targets into one


thomas-fossati:
We need to merge the "extra" lint target into the "normal" one as we've have done for other packages.
All lint errors must be addressed, either changing the code or suppressing them (when there's a good reason to do so).

PR #153, opened 2025-01-13
Merge "extra" lint target into "normal" lint target

Akhilesh-max:
Fixes -- #91 

This PR merges the "extra" lint target into the "normal" lint target in the Makefile.


Akhilesh-max:
Hi @thomas-fossati, I have merged the "extra" lint into the "normal" lint target. Please take a look when get a chance.

Akhilesh-max:
Hi @thomas-fossati, I have merged the "extra" lint into the "normal" lint target. Most of the errors have been fixed, and some have been suppressed wherein reasonable to do so. I have also updated the workflow file. PTAL when you get a chance.

Thanks!!

cots/test_vars.go
```diff
@@ -42,15 +43,15 @@ var (
 				{
 					"format": 0,
 					"data": "MIIBvTCCAWSgAwIBAgIVANCdkL89UlzHc9Ui7XfVniK7pFuIMAoGCCqGSM49BAMCMD4xCzAJBgNVBAYMAlVTMRAwDgYDVQQKDAdFeGFtcGxlMR0wGwYDVQQDDBRFeGFtcGxlIFRydXN0IEFuY2hvcjAeFw0yMjA1MTkxNTEzMDdaFw0zMjA1MTYxNTEzMDdaMD4xCzAJBgNVBAYMAlVTMRAwDgYDVQQKDAdFeGFtcGxlMR0wGwYDVQQDDBRFeGFtcGxlIFRydXN0IEFuY2hvcjBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABONRqhA5JAekvQN8oLwRVNDnAfBnTznLLE+SEGks677sHSeXfcVhZXUeDiN7/fsVNumaiEWRQpZh3zXPwL8rUMyjPzA9MB0GA1UdDgQWBBQBXEXJrLBGKnFd1xCgeMAVSfEBPzALBgNVHQ8EBAMCAoQwDwYDVR0TAQH/BAUwAwEB/zAKBggqhkjOPQQDAgNHADBEAiALBidABsfpzG0lTL9Eh9b6AUbqnzF+koEZbgvppvvt9QIgVoE+bhEN0j6wSPzePjLrEdD+PEgyjHJ5rbA11SPq/1M="
-				},
+				},
```
setrofim, line 46: Is the trailing space intentional?

cots/test_vars.go
```diff
@@ -42,15 +43,15 @@ var (
...
 				},
 				{
 					"format": 1,
 					"data": "ooIC1TCCAtEwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAATN0f5kzywEzZOYbaV23O3N8cku39JoLNjlHPwECbXDDWp0LpAO1z248/hoy6UW/TZMTPPR/93XwHsG16mSFy8XBBSKhM/5gJWjvDbW7qUY1peNm9cfYDCCAlwwXDELMAkGA1UEBgwCVVMxHzAdBgNVBAoMFlNub2JiaXNoIEFwcGFyZWwsIEluYy4xLDAqBgNVBAMMI1Nub2JiaXNoIEFwcGFyZWwsIEluYy4gVHJ1c3QgQW5jaG9yoIIB+jCCAZ+gAwIBAgIUEBuTRGXAEEVEHhu4xafAnqm+qYgwCgYIKoZIzj0EAwIwXDELMAkGA1UEBgwCVVMxHzAdBgNVBAoMFlNub2JiaXNoIEFwcGFyZWwsIEluYy4xLDAqBgNVBAMMI1Nub2JiaXNoIEFwcGFyZWwsIEluYy4gVHJ1c3QgQW5jaG9yMB4XDTIyMDUxOTE1MTMwOFoXDTMyMDUxNjE1MTMwOFowXDELMAkGA1UEBgwCVVMxHzAdBgNVBAoMFlNub2JiaXNoIEFwcGFyZWwsIEluYy4xLDAqBgNVBAMMI1Nub2JiaXNoIEFwcGFyZWwsIEluYy4gVHJ1c3QgQW5jaG9yMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEzdH+ZM8sBM2TmG2ldtztzfHJLt/SaCzY5Rz8BAm1ww1qdC6QDtc9uPP4aMulFv02TEzz0f/d18B7BtepkhcvF6M/MD0wHQYDVR0OBBYEFIqEz/mAlaO8NtbupRjWl42b1x9gMAsGA1UdDwQEAwIChDAPBgNVHRMBAf8EBTADAQH/MAoGCCqGSM49BAMCA0kAMEYCIQC2cf43f3PPlCO6/dxv40ftIgxxToKHF72UzENv7+y4ygIhAIGtC/r6SGaFMaP7zD2EloBuIXTtyWu8Hwl+YGdXRY93"
-				}
+				}
```
setrofim, line 54: Is the trailing space intentional?

Akhilesh-max: Nope, removed.

Akhilesh-max:
I’ve made the suggested changes. Please review them when get a chance. @setrofim, @yogeshbdeshpande, @thomas-fossati 

Thanks!

setrofim:
Please could you also squash the fixup commits.

yogeshbdeshpande:
Thanks @Akhilesh-max : Reviewing it now but in the mean time, please follow the squash advise from Sergei

comid/example_psa_refval_test.go
```diff
@@ -69,15 +69,16 @@ func extractSwMeasurements(m Measurements) error {
 	if len(m.Values) == 0 {
 		return fmt.Errorf("no measurements")
 	}
-	for i, m := range m.Values {
-		if err := extractSwMeasurement(m); err != nil {
+	for i := range m.Values {
+		meas := &m.Values[i]
```
yogeshbdeshpande, line 73: traditionally golang always follows conciseness!

Wondering what lint warning is thrown for this change, I mean is line 73 change this really needed ..?

Could you please clarify.

yogeshbdeshpande (changes requested):
Thanks for the changes, I have reviewed it: Most of the changes looks good to me, except one question I put forth in the Review

comid/example_psa_refval_test.go
thomas-fossati: The old L72 makes a copy of `m` at each iteration.
The new L72 takes a reference instead.

yogeshbdeshpande: Then should not it be better to set extractSwMeasurement(&m)
Or if copy needs to be avoided, then 

extractSwMeasurement(&meas.Values[i])

thomas-fossati: > Then should not it be better to set extractSwMeasurement(&m)

hmm no, that changes nothing wrt the copy.  It's already happened

> Or if copy needs to be avoided, then
> 
> extractSwMeasurement(&meas.Values[i])

I guess you mean `extractSwMeasurement(&m.Values[i])`?
If so, it'd be equivalent to the proposed change, except it's a bit less readable.

comid/comid.go
```diff
@@ -242,7 +242,7 @@ func (o *Comid) AddDevIdentityKey(val KeyTriple) *Comid {
 	return o
 }
 
-func (o Comid) Valid() error {
+func (o *Comid) Valid() error {
```
thomas-fossati, line 245: `Valid` does not modify the `Comid` object.

Therefore, pass-by-reference conveys the wrong semantics.

I guess this is gocritic's `hugeParam` check?  If so, we might as well silence it.
(Maybe even remove it altogether from the linter config.)

comid/comid.go
```diff
@@ -306,18 +306,18 @@ func (o Comid) ToJSON() ([]byte, error) {
...
+	return encoding.SerializeStructToJSON(o)
 }
 
 // FromJSON deserializes a JSON-encoded CoMID into the target Comid
 func (o *Comid) FromJSON(data []byte) error {
 	return encoding.PopulateStructFromJSON(data, o)
 }
 
-func (o Comid) ToJSONPretty(indent string) ([]byte, error) {
+func (o *Comid) ToJSONPretty(indent string) ([]byte, error) {
```
thomas-fossati, line 317: ditto (see https://github.com/veraison/corim/pull/153/files#r1914729206)

There are other instances of this.

comid/example_psa_refval_test.go
thomas-fossati: > > > Then should not it be better to set extractSwMeasurement(&m)
> > 
> > 
> > hmm no, that changes nothing wrt the copy. It's already happened
> 
> True that , however now we copy a pointer instead of a value, how different then it is to the original change ?

In yours, the copy has already happened in the `for` statement, you are not avoiding it by taking the reference to `m`: it's too late.
In the proposed change, the measurement object is not copied.

> > > Or if copy needs to be avoided, then
> > > extractSwMeasurement(&meas.Values[i])
> > 
> > 
> > I guess you mean `extractSwMeasurement(&m.Values[i])`? If so, it'd be equivalent to the proposed change, except it's a bit less readable.
> 
> ok, but this avoids one extra variable addition, so much neater, in my view!

Readability is, by definition, in the eye of the reader :-)
Personally, I do not like cramming too much information into a single line, therefore I prefer the proposed change.
Anyway, this is not a hill I'd die on :-) I'll let @Akhilesh-max decide.

comid/example_test.go
```diff
@@ -109,8 +109,8 @@ func Example_encode() {
...
...41434d45204c74642e026a526f616452756e6e65720300040101d902264702deadbeefdead02d8255031fb5abf023e4992aa4e95f9c1503bfa81a200d8255031fb5abf023e4992aa4e95f9c1503bfa01aa01d90229020282820644abcdef00820644ffffffff03a300f401f403f504d9023044010203040544ffffffff064802005e1000000001075020010db8000000000000000000000068086c43303258373056484a484435094702deadbeefdead0a5031fb5abf023e4992aa4e95f9c1503bfa028182a101d902264702deadbeefdead81d9022a78b12d2d2d2d2d424547494e205055424c4943204b45592d2d2d2d2d0a4d466b77457759484b6f5a497a6a3043415159494b6f5a497a6a304441516344516741455731427671462b2f727938425761375a454d553178595948455138420a6c4c54344d46484f614f2b4943547449767245654570722f7366544150363648326843486462354845584b74524b6f6436514c634f4c504131513d3d0a2d2d2d2d2d454e44205055424c4943204b45592d2d2d2d2d038182a101d8255031fb5abf023e4992aa4e95f9c1503bfa81d9022a78b12d2d2d2d2d424547494e205055424c4943204b45592d2d2d2d2d0a4d466b77457759484b6f5a497a6a3043415159494b6f5a497a6a304441516344516741455731427671462b2f727938425761375a454d553178595948455138420a6c4c54344d46484f614f2b4943547449767245654570722f7366544150363648326843486462354845584b74524b6f6436514c634f4c504131513d3d0a2d2d2d2d2d454e44205055424c4943204b45592d2d2d2d2d
```
thomas-fossati, line 112: these seem gratuitous changes.  If it's another gocritic lamentation we may want to suppress it.

comid/comid.go
setrofim: Will that not result in unnecessary  copies of a large object? 

I suppose, we shouldn't worry about perf until this can be shown to be a problem, but univerasally disabling the flag makes me feel a bit uneasy. (Makes me wish Go had mutable and immutable references a la Rust...)

setrofim:
Sorry to nit-pick, but could you also ammend the commit message to adhere to conventional commits practice and to be more informative. Something like

```
chore(make): merge extra-lint target into lint

Add extra-lint checks to lint target to ensure they are always run during CI,
and remove the need for an additional target that needs to be executed manually
(which is often neglected resulting in gradual accumulation of regressions on
main).
```

comid/comid.go
thomas-fossati: > Will that not result in unnecessary copies of a large object?

Yes, that is likely the thing we are warned about here.

> I suppose, we shouldn't worry about perf until this can be shown to be a problem, but univerasally disabling the flag makes me feel a bit uneasy.

I'm OK with a local lint annotation.

Tangentially, I think we may want to also take a critical look at the checks we are enabling and whether they are the right set.

> (Makes me wish Go had mutable and immutable references a la Rust...)

Grust FTW! :-D

setrofim: > Tangentially, I think we may want to also take a critical look at the checks we are enabling and whether they are the right set.

Agreed, though I do think `hugeParam` check is a reasonable thing to have enabled (I do remember it being useful thing to me in the past when unintentionally copying a larger structure than CoRIM). 

> Grust FTW! :-D

Lol :D.  I'd love to see a more cargo-like build/dependency management system for Go as well...

comid/example_psa_refval_test.go
yogeshbdeshpande: Honestly, i did not like the suggested change from lint-extra, it kind of defeats the purpose of how effectively one uses golang `for loop`, to have access to its index and its variables.

comid/comid.go
Akhilesh-max: @setrofim @thomas-fossati, via the conversation above, I am able to deduce two things --

i) We will use local lint annotations to silence some of the errors where it is rational, without disabling the hugeParam check. Shall I silence all the hugeParam linting errors on value receivers in the context of this [comment](https://github.com/veraison/corim/pull/153#discussion_r1914730064).

ii) We need to review the enabled linter checks. (Created an issue for this - #155 )

comid/comid.go
Akhilesh-max: Sure, would suppress the hugeParam check herein. 
However, please suggest to the query in this [comment](https://github.com/veraison/corim/pull/153/files#r1916296214).

comid/example_test.go
Akhilesh-max: Added a space between // and the comment text. This should be fine?

Akhilesh-max:
@setrofim Squashed the fixup commits and amended the commit message. Thanks for the suggestion though :).

comid/comid.go
setrofim: i) Yes, please make that change. (Personally, I think what you have now is fine; however as Thomas wants to preserve the semantics of pointer receivers implying modification in the method, I'm onboard with suppressing the warning for now -- we don't care too much about perf at the moment, and the `Comid` structure is not massive -- I estimate ~90 bytes -- so a few additional copies is not a big deal).

ii) Thank you!

Akhilesh-max: OK -- We have approximately 26 hugeParam linting errors on value receivers. Below are the structs, their sizes, and the number of occurrences:
	•	EatCWTClaim: 176 bytes, 2 occurrences
	•	conciseTaStore: 120 bytes, 3 occurrences
	•	AbbreviatedSwidTag: 180 bytes, 4 occurrences
	•	unsignedCorim: 88 bytes, 4 occurrences
	•	Measurement: 128 bytes, 2 occurrences
	•	Mval: 112 bytes, 3 occurrences
	•	Flagsmap: 88 bytes, 4 occurrences
	•	Comid: 112 bytes, 4 occurrences

Is it OK to suppress hugeParam for all warnings?

setrofim: Yeah, that all seems fine. Thanks for checking.

setrofim (approved):
LGTM

Akhilesh-max:
@yogeshbdeshpande @thomas-fossati PTAL.
Thanks!

yogeshbdeshpande (approved):
LGTM!

thomas-fossati (approved):
Thank you very much!

PR #156, opened 2025-01-19
comid/environment: improve ToCBOR and FromCBOR tests

pranjalkole:
The ToCBOR and FromCBOR have been changed to the following order:
Empty
Class only
Instance only
Group only
Class and Instance
Class and Group
Instance and Group
Class and Instance and Group

The missing tests have also been added.


thomas-fossati (approved):
Thank you @pranjalkole !

PR #157, opened 2025-01-19
corim/signedcorim: check for mandatory alg and kid

pranjalkole:
Before merging, we need to change `corim/testcases/signed-good-corim.cbor` to include the `kid` parameter.


corim/signedcorim.go
```diff
@@ -68,9 +73,10 @@ func (o *SignedCorim) processHdrs() error {
 		return fmt.Errorf("expecting content type %q, got %q instead", ContentType, v)
 	}
 
-	// TODO(tho) key id is apparently mandatory, which doesn't look right.
-	// TODO(tho) Check with the CoRIM design team.
-	// See https://github.com/veraison/corim/issues/14
+	v, ok = hdr.Protected[cose.HeaderLabelKeyID]
+	if !ok {
+//		return errors.New("missing mandatory key id")
```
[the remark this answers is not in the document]
thomas-fossati: I suspect `kid` optionality is backed into some of the existing tests.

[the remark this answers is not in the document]
setrofim: Yeah, I expect this change would require more tests to be fixed up than just the test case mentioned.

corim/signedcorim.go
```diff
@@ -159,6 +179,7 @@ func (o *SignedCorim) Sign(signer cose.Signer) ([]byte, error) {
 
 	o.message.Headers.Protected.SetAlgorithm(alg)
 	o.message.Headers.Protected[cose.HeaderLabelContentType] = ContentType
+	o.message.Headers.Protected[cose.HeaderLabelKeyID] = kid
```
thomas-fossati, line 182: we should we ensure that `kid != nil`

pranjalkole: `kid` is supposed to never be `nil`. I was thinking we should store `kid` as part of the `cose.Signer` while creating it from the JWK (like we do for `alg`). `getKidFromJWK` can get a `kid` from every `JWK`.

I'll wait for the meeting on `kid` being mandatory before making any more changes.

pranjalkole: @thomas-fossati Please ping me after the final decision on whether `kid` should be mandatory.

Issue #164, opened 2025-02-07
Create shell script to synthesize a test cert chain


thomas-fossati:
Repurpose [the RATSd script](https://github.com/veraison/ratsd/blob/main/gen-certs) to create a 3-level chain including:

1. root CA certificate with subject name "Acme Inc."
2. intermediate CA cert with subject name "Acme Gizmos"
3. an end-entity cert with subject name "Acme Gizmo CoRIM signer"

The name are expressed purely as CNs (i.e., no SAN).

The synthesized chain will be stored in a new `misc` folder under the `veraison/corim` repo.
The script will be stored in the `scripts` folder under the `veraison/corim` repo.

Also, add a new target to the top-level Makefile to regenerate the chain.

yogeshbdeshpande:
@thomas-fossati is there anything left to be done with this issue, else, please close this...

PR #165, opened 2025-02-10
feat: adds 3-level certificate chain generation script

Akhilesh-max:
Fixes -- #164


Akhilesh-max:
@thomas-fossati Please take a look when get a chance. 
Thanks!

scripts/gen-certs.sh
```diff
@@ -0,0 +1,161 @@
...
+    help
+            Print this message and exit (same as -h option).
+
+    Options:
+
+    -h         Print this message and exit.
+    -C         Do not clean up intermediate artifacts (e.g., CSRs).
+
+EOF
+
```
setrofim, line 106: `set -e` here to re-enable fail-on-error after the `read`. It doesn't matter too much, as you're gonna exist straight after calling `help` anywya, but still good to have here for robustness, in case things change in the future.

Makefile
```diff
@@ -49,6 +49,11 @@ presubmit:
 .PHONY: licenses
 licenses: ; @./scripts/licenses.sh
 
+.PHONY: certs
```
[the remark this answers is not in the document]
thomas-fossati: When scoping this task, @Akhilesh-max and I decided it'd be handy to have a target to regenerate the chain when needed (e.g., when certs expire).

See https://github.com/veraison/corim/issues/164

I think it should be made clearer in the help that regenerating the chain is an exceptional action.

scripts/gen-certs.sh
thomas-fossati, line 1: Please, break long lines.  (Ideally at 79.)

scripts/gen-certs.sh
```diff
@@ -0,0 +1,161 @@
+#!/bin/bash
+# SPDX-License-Identifier: Apache-2.0
+set -e
```
thomas-fossati, line 3: We want to fail hard if there are unset variables.

Besides, it's good practice to pair `-e` with `set -o pipefail` to catch errors in command pipelines.  There are none here that I can see at the moment, but it costs little and would make the script more robust if `|`s are added in the future.

```suggestion
set -eu
set -o pipefail
```

scripts/gen-certs.sh
thomas-fossati, line 1: Since we exit on error, and that could happen anywhere, we may want to set an EXIT trap to clean up any leftovers (i.e., a `trap clean_intermediate EXIT`.)

One could argue that the really clean way to deal with this would be removing all files (including certs and keys) and restoring their state from git.
I am not sure we want to go that far for what is essentially a one-off script.

scripts/gen-certs.sh
```diff
@@ -0,0 +1,161 @@
...
+    fi
+    openssl ecparam -name prime256v1 -genkey -noout -out ${MISC_DIR}/${END_ENTITY_CERT_NAME}.key
+    openssl req -new -key ${MISC_DIR}/${END_ENTITY_CERT_NAME}.key -out ${MISC_DIR}/${END_ENTITY_CERT_NAME}.csr -subj "/CN=Acme Gizmo CoRIM signer"
+    openssl x509 -req -in ${MISC_DIR}/${END_ENTITY_CERT_NAME}.csr -CA ${MISC_DIR}/${INTERMEDIATE_CERT_NAME}.der -CAkey ${MISC_DIR}/${INTERMEDIATE_CERT_NAME}.key -CAcreateserial -out ${MISC_DIR}/${END_ENTITY_CERT_NAME}.crt -days 825 -sha256 -CAform der
+    openssl x509 -in ${MISC_DIR}/${END_ENTITY_CERT_NAME}.crt -outform der -out ${MISC_DIR}/${END_ENTITY_CERT_NAME}.der
+    rm -f ${MISC_DIR}/${END_ENTITY_CERT_NAME}.crt  # Remove the PEM certificate
+    echo "Created ${MISC_DIR}/${END_ENTITY_CERT_NAME}.der and ${MISC_DIR}/${END_ENTITY_CERT_NAME}.key"
+}
+
+function clean_intermediate() {
```
thomas-fossati, line 53: Quibble: maybe we can find a better name than "intermediate" to avoid confusion with "intermediate CA".

thomas-fossati (commented):
Good start!  I have left a few comments inline.

scripts/gen-certs.sh
setrofim: Isn't that what the intermediate certs supposed to represent?

Makefile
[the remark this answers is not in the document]
Akhilesh-max: Added a note in the help to mention the same.

scripts/gen-certs.sh
Akhilesh-max: Done.

Akhilesh-max: Thanks for the suggestions.

I am not sure either, however maybe restoring from git could be an overkill for our use case since the script runs so infrequently, mainly when certificates expire. And, when the certs would have expired, we won't be having any desired state. Thus, using clean_all/manual cleanup should be sufficient to ensure a clean state, am I right?

I’ve added an EXIT trap to call clean_certs_artifacts, if -C flag is not passed.

scripts/gen-certs.sh
Akhilesh-max: Added

scripts/gen-certs.sh
```diff
@@ -0,0 +1,161 @@
...
+}
+
+function create_end_entity_cert() {
+    if ([[ -f "${MISC_DIR}/${END_ENTITY_CERT_NAME}.der" ]] && [[ -f "${MISC_DIR}/${END_ENTITY_CERT_NAME}.key" ]]); then
+        echo "End-entity certificate and key already exist. Skipping creation."
+        return
+    fi
+    openssl ecparam -name prime256v1 -genkey -noout -out ${MISC_DIR}/${END_ENTITY_CERT_NAME}.key
+    openssl req -new -key ${MISC_DIR}/${END_ENTITY_CERT_NAME}.key -out ${MISC_DIR}/${END_ENTITY_CERT_NAME}.csr -subj "/CN=Acme Gizmo CoRIM signer"
+    openssl x509 -req -in ${MISC_DIR}/${END_ENTITY_CERT_NAME}.csr -CA ${MISC_DIR}/${INTERMEDIATE_CERT_NAME}.der -CAkey ${MISC_DIR}/${INTERMEDIATE_CERT_NAME}.key -CAcreateserial -out ${MISC_DIR}/${END_ENTITY_CERT_NAME}.crt -days 825 -sha256 -CAform der
```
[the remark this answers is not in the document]
Akhilesh-max: Done

scripts/gen-certs.sh
Akhilesh-max: Thanks for the suggestions. Included.

scripts/gen-certs.sh
Akhilesh-max: Updated to -- clean_certs_artifacts

Akhilesh-max:
@thomas-fossati @setrofim Addressed all the comments above. PTAL.
Thanks!

Makefile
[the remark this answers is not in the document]
thomas-fossati: > Added a note in the help to mention the same.

Great.  However, I can't see that in [2d5c5c3](https://github.com/veraison/corim/pull/165/commits/2d5c5c31305e1ba2b481df149cc82510763ffa81), did you forget to commit?

scripts/gen-certs.sh
thomas-fossati: Sorry to be a pest, but there is still one long line (around L30) to split ;-)

scripts/gen-certs.sh
thomas-fossati: > Isn't that what the intermediate certs supposed to represent?

Yes.  This function deals with CSRs and other byproducts rather than intermediate CA certs.

scripts/gen-certs.sh
```diff
@@ -0,0 +1,202 @@
...
+    help
+            Print this message and exit (same as -h option).
+
+    Options:
+
+    -h         Print this message and exit.
+    -C         Do not clean up intermediate artifacts (e.g., CSRs).
+
+    Note: Regenerating the certificate chain is an exceptional action and should
+    only be done when necessary (e.g., when certificates expire).
```
Akhilesh-max, line 143: @thomas-fossati 
wrt https://github.com/veraison/corim/pull/165/files#r1951084868
Is it the way we wanted it?

Makefile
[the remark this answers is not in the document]
setrofim: IMO this still needs to move out of the top-level make file, or at leat, rename the target to something like `test-cert`s to make it clear that this is not necessary to build or use the library and is only to (re)generate test cases.

scripts/gen-certs.sh
Akhilesh-max: Sorry to miss that. Done.

Makefile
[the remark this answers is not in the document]
thomas-fossati: I like the `test-cert` target idea.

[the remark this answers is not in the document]
Akhilesh-max: `test-certs`?

thomas-fossati (approved):
Thanks for addressing my comments.

🚢 it!

Akhilesh-max:
@setrofim PTAL. Thanks!

scripts/gen-certs.sh
```diff
@@ -0,0 +1,202 @@
...
+    deployment. The certificates are signed by a CA certificate called
+    ${ROOT_CERT_NAME}.crt. If this does not exist, a self-signed one will
+    be generated.
+
+    Commands:
+
+    create
+            Create the root, intermediate, and end-entity certificates.
+
+    clean
```
yogeshbdeshpande, line 127: It is not clear the scope of `clean` to me?

Is it all certs, then please specify clearly..

Akhilesh-max: Done.

yogeshbdeshpande (commented):
Most of my comments have been seeking clarifications and minor .nits.

I will approve once you please clarify!

scripts/gen-certs.sh
Akhilesh-max: It should have been clean_certs_artifacts. Updated accordingly.

Makefile
```diff
@@ -58,3 +63,4 @@ help:
 	@echo "  * presubmit:  check you are ready to push your local branch to remote"
 	@echo "  * help:       print this menu"
 	@echo "  * licenses:   check licenses of dependent packages"
+	@echo "  * test-certs:      generate the certificate chain"
```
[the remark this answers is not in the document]
Akhilesh-max: Done.

scripts/gen-certs.sh
Akhilesh-max: Done.

Akhilesh-max:
@yogeshbdeshpande @setrofim Updated as per the reviews.

yogeshbdeshpande (approved):
LGTM! Thanks for the changes!

setrofim (approved):

PR #167, opened 2025-02-11
Add routines to compare measurements against reference

jraman567:
Adds a compare routine to some of the measurement types.

### DIFF
--- a/comid/digests.go
+++ b/comid/digests.go
@@ -1,9 +1,10 @@
-// Copyright 2021 Contributors to the Veraison project.
+// Copyright 2021-2025 Contributors to the Veraison project.
 // SPDX-License-Identifier: Apache-2.0
 
 package comid
 
 import (
+	"bytes"
 	"fmt"
 
 	"github.com/veraison/swid"
@@ -39,6 +40,86 @@ func (o Digests) Valid() error {
 		}
 	}
 	return nil
+}
+
+// Equal confirms if the Digests object is the same as another
+//
+// Two digests are considered to be equal if they meet the following criteria:
+//   - They contain the same number of elements
+//   - All the elements that use the same algorithm have the same value,
+//     though the elements could be in any order
+func (o Digests) Equal(r Digests) bool {
+	om := make(map[uint64][][]byte)
+	for _, oe := range o {
+		vs, ok := om[oe.HashAlgID]
+		if ok {
+			om[oe.HashAlgID] = append(vs, oe.HashValue)
+		} else {
+			om[oe.HashAlgID] = [][]byte{oe.HashValue}
+		}
+	}
+
+outer:
+	for _, re := range r {
+		ovs, ok := om[re.HashAlgID]
+		if !ok {
+			return false
+		}
+
+		for _, ov := range ovs {
+			if bytes.Equal(ov, re.HashValue) {
+				continue outer
+			}
+		}
+
+		return false
+	}
+
+	return true
+}
+
+// CompareAgainstReference checks if digests object matches with a reference
+//
+//	See the following CoRIM spec for rules to compare
+//	digests against a reference:
+//	https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#section-8.5.6.1.3
+func (o Digests) CompareAgainstReference(r Digests) bool {
+	result := false
+
+	if len(r) == 0 {
+		return false
+	}
+
+	// Insert the reference values into a map
+	ref := make(map[uint64][]byte)
+	for _, digest := range r {
+		val, ok := ref[digest.HashAlgID]
+		if ok && !bytes.Equal(digest.HashValue, val) {
+			// If two entries with the same hashing algorithm have different
+			// values, that's an automatic false.
+			return false
+		}
+
+		ref[digest.HashAlgID] = digest.HashValue
+	}
+
+	// Check the object against the reference value map
+	for _, digest := range o {
+		val, ok := ref[digest.HashAlgID]
+		if !ok {
+			continue
+		}
+
+		if !bytes.Equal(digest.HashValue, val) {
+			// All hash values must be equal if a claim has the same
+			// digest represented using multiple algorithms.
+			return false
+		}
+
+		result = true
+	}
+
+	return result
 }
 
 func NewHashEntry(algID uint64, value []byte) *swid.HashEntry {
--- a/comid/flagsmap.go
+++ b/comid/flagsmap.go
@@ -1,10 +1,11 @@
-// Copyright 2023-2024 Contributors to the Veraison project.
+// Copyright 2023-2025 Contributors to the Veraison project.
 // SPDX-License-Identifier: Apache-2.0
 
 package comid
 
 import (
 	"fmt"
+	"reflect"
 
 	"github.com/veraison/corim/encoding"
 	"github.com/veraison/corim/extensions"
@@ -177,6 +178,14 @@ func (o *FlagsMap) Get(flag Flag) *bool {
 	default:
 		return o.Extensions.get(flag)
 	}
+}
+
+func (o FlagsMap) Equal(r FlagsMap) bool { //nolint:gocritic
+	return reflect.DeepEqual(o, r)
+}
+
+func (o FlagsMap) CompareAgainstReference(r FlagsMap) bool { //nolint:gocritic
+	return o.Equal(r)
 }
 
 // RegisterExtensions registers a struct as a collections of extensions
--- a/comid/integrityregisters.go
+++ b/comid/integrityregisters.go
@@ -1,10 +1,11 @@
-// Copyright 2024 Contributors to the Veraison project.
+// Copyright 2024-2025 Contributors to the Veraison project.
 // SPDX-License-Identifier: Apache-2.0
 package comid
 
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"strconv"
 
 	"github.com/veraison/swid"
@@ -135,4 +136,33 @@ func (i *IntegrityRegisters) UnmarshalJSON(data []byte) error {
 		}
 	}
 	return nil
+}
+
+func (i IntegrityRegisters) Equal(r IntegrityRegisters) bool {
+	return reflect.DeepEqual(i, r)
+}
+
+// CompareAgainstReference checks if IntegrityRegisters object matches with a reference
+//
+//	See the following CoRIM spec for rules to compare
+//	IntegrityRegisters against a reference:
+//	https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#name-comparison-for-integrity-re
+func (i IntegrityRegisters) CompareAgainstReference(r IntegrityRegisters) bool {
+	result := false
+
+	for refIndex, refDigests := range r.IndexMap {
+		claimDigests, ok := i.IndexMap[refIndex]
+		if !ok {
+			return false
+		}
+
+		ref := refDigests
+		if !claimDigests.CompareAgainstReference(ref) {
+			return false
+		}
+
+		result = true
+	}
+
+	return result
 }
--- a/comid/macaddr.go
+++ b/comid/macaddr.go
@@ -1,4 +1,4 @@
-// Copyright 2021-2024 Contributors to the Veraison project.
+// Copyright 2021-2025 Contributors to the Veraison project.
 // SPDX-License-Identifier: Apache-2.0
 
 package comid
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"reflect"
 )
 
 // MACaddr is an HW address (e.g., IEEE 802 MAC-48, EUI-48, EUI-64)
@@ -15,6 +16,14 @@ import (
 // (see: https://github.com/golang/go/issues/29678)
 // we need to create an alias type with a custom decoder.
 type MACaddr net.HardwareAddr
+
+func (o MACaddr) Equal(r MACaddr) bool {
+	return reflect.DeepEqual(o, r)
+}
+
+func (o MACaddr) CompareAgainstReference(r MACaddr) bool {
+	return o.Equal(r)
+}
 
 // UnmarshalJSON deserialize a MAC address in textual form into the MACaddr
 // target, e.g.:
--- a/comid/rawvalue.go
+++ b/comid/rawvalue.go
@@ -1,11 +1,14 @@
-// Copyright 2021-2024 Contributors to the Veraison project.
+// Copyright 2021-2025 Contributors to the Veraison project.
 // SPDX-License-Identifier: Apache-2.0
 
 package comid
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
+	"log"
+	"reflect"
 )
 
 // RawValue models a $raw-value-type-choice.  For now, the only available type is bytes.
@@ -39,6 +42,36 @@ func (o RawValue) GetBytes() ([]byte, error) {
 	default:
 		return nil, fmt.Errorf("unknown type %T for $raw-value-type-choice", t)
 	}
+}
+
+func (o RawValue) Equal(r RawValue) bool {
+	return reflect.DeepEqual(o, r)
+}
+
+func (o RawValue) CompareAgainstReference(ref []byte, mask *[]byte) bool {
+	claim, err := o.GetBytes()
+	if err != nil {
+		log.Printf("RawValue:CompareAgainstReference: Error: %v", err)
+		return false
+	}
+
+	if mask != nil {
+		if len(claim) != len(ref) {
+			return false
+		}
+
+		if len(*mask) != len(claim) {
+			log.Printf("RawValue:CompareAgainstReference: Error: mask length")
+			return false
+		}
+
+		for i := range *mask {
+			claim[i] = (*mask)[i] & claim[i]
+			ref[i] = (*mask)[i] & ref[i]
+		}
+	}
+
+	return bytes.Equal(claim, ref)
 }
 
 func (o RawValue) MarshalCBOR() ([]byte, error) {
--- a/comid/svn.go
+++ b/comid/svn.go
@@ -1,4 +1,4 @@
-// Copyright 2021-2024 Contributors to the Veraison project.
+// Copyright 2021-2025 Contributors to the Veraison project.
 // SPDX-License-Identifier: Apache-2.0
 
 package comid
@@ -6,6 +6,7 @@ package comid
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"strconv"
 
 	"github.com/veraison/corim/encoding"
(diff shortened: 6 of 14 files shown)


jraman567:
I can't figure out the lint error; I'm unsure how to fix it.

thomas-fossati:
> I can't figure out the lint error; I'm unsure how to fix it.

I suggest giving a go at:
```console
goimports -local "github.com/veraison/corim" -w .
```

jraman567:
> I suggest giving a go at:
> 
> ```
> goimports -local "github.com/veraison/corim" -w .
> ```

Thanks, @thomas-fossati; that helped to fix it. As Dionna suggested in the meeting, the fix needed a new line between "fmt" and the following line.

I'm working on the test coverage and will post a new revision of the changes shortly. Thank you!

comid/flagsmap.go
```diff
@@ -179,6 +180,24 @@ func (o *FlagsMap) Get(flag Flag) *bool {
 	}
 }
 
+func (o *FlagsMap) Compare(r *FlagsMap) error {
+	claim, err := o.MarshalJSON()
```
setrofim, line 184: JSON conversion seems to be  an unncessarily round-about way of doing this. Also, doing byte comparisons of JSON is always iffy. I think this is _probably_ fine in this case, however, if `Extensions` happen to contain maps for some reason, they may not always serilize the same way.

comid/integrityregisters.go
```diff
@@ -136,3 +137,21 @@ func (i *IntegrityRegisters) UnmarshalJSON(data []byte) error {
 	}
 	return nil
 }
+
+func (i *IntegrityRegisters) Compare(r *IntegrityRegisters) error {
+	claim, err := i.MarshalJSON()
```
setrofim, line 142: Same comment as above, and in this case, since `IntegrityRegisters` has a map, byte comparison of JSON is _definitely_ not safe.

comid/macaddr.go
```diff
@@ -16,6 +17,24 @@ import (
 // we need to create an alias type with a custom decoder.
 type MACaddr net.HardwareAddr
 
+func (o *MACaddr) Compare(r *MACaddr) error {
+	claim, err := o.MarshalJSON()
```
setrofim, line 21: Ditto.

setrofim (changes requested):
As a general note, it's weird for `Compare()` method to raise an error. Two values not being equal is not automatically considered an error condition. Also, I would expect a method called "Compare" to do a relative comparison, rather than equality test.

I would rename the method to `Equal` and have it return a `bool` instead.

 (Alternatively, you could keep  it as "Compare" and have it return an `int`, with `1` indicating  `o > r`, `-1`  indicating `r > o`, and `0` indicating `o == r`; however, that would require chainging the implementations as well, and would require for all types in question to have partial ordering, which I don't think we want.)

thomas-fossati:
> Thanks, @thomas-fossati; that helped to fix it. As Dionna suggested in the meeting, the fix needed a new line between "fmt" and the following line.

Do you use vim or VS Code?  If so, the popular go plugins (e.g., vim-go) should automatically update the import statements order transparently.

thomas-fossati (commented):
I agree with @setrofim high-level comment.

I think we should have:
* a `Compare` method for types for which ordering makes sense, e.g., SVNs -- and it should return -1,0,1 rather than error;
* an `Equal` method for types for which we are only concerned about strict equality (e.g., digests) -- and this could return a bool.

setrofim:
> an Equal method for types for which we are only concerned about strict equality (e.g., digests)

I think we should have `Equal` for the other types as well (even if it's just implemented as `o.Compare(r) ==  0`, since that would allow you to handle of values uninformly as `interface{ Equal() bool }` in situatations where you only care about exact values.

thomas-fossati:
> > an Equal method for types for which we are only concerned about strict equality (e.g., digests)
> 
> I think we should have `Equal` for the other types as well (even if it's just implemented as `o.Compare(r) == 0`), since that would allow you to handle of values uninformly as `interface{ Equal() bool }` in situatations where you only care about exact values.

WFM

jraman567:
OK, I'll push a PR with recommended changes. Thanks @thomas-fossati & @setrofim for the review.

comid/svn.go
```diff
@@ -182,6 +203,41 @@ func (o TaggedMinSVN) Valid() error {
...
 }
 
+// Compare compares SVN to a reference
+//
+// returns:
+//
+// 0 if they are equal
+// 1 if SVN is newer than the reference
+// -1 if SVN is older than reference or if the method encounters an error
+func (o TaggedMinSVN) Compare(r TaggedMinSVN) int {
```
thomas-fossati, line 213: I'm wondering about the semantics of this.  I was expecting to compare a "normal" SVN against a "minSVN", no?  What does it mean to compare two "minSVN"s?

comid/svn.go
```diff
@@ -182,6 +203,41 @@ func (o TaggedMinSVN) Valid() error {
...
+	} else if claim > ref {
+		return 1
+	}
+
+	return 0
+}
+
+// Equal Semantics for equality of TaggedMinSVN mean that the claim
+// is the same or newer than the reference
+func (o TaggedMinSVN) Equal(r TaggedMinSVN) bool {
```
setrofim, line 237: This... is weird. It's conflating matching semantics with equality.  Also, as @thomas-fossati stated above, in the context of verification, minSVN would have an SVN compared against it, not another minSVN.  And, mathematically speaking, one of the properties of equality is symmetry, which this does not have...

So IMO either this method should be either fixed to represent equality, i.e. `o.Compare(r) == 0`, or removed entirely if we decide that comarison between two minSVN's never makes sense.

comid/svn.go
setrofim: That's true for verification, but in other contexts it might make sense? E.g. if you're comparing two policies. Also, changing the arument type would make the semantics for this `Compare` slightly different from the others (which compare like values, where as this compares against a threshold).

Maybe it would make sense to add a different method, `ValidateSVN(v TaggedSVN) bool` or something, for valiation (and a mirroring `CompareAgainstMinSVN(v TaggedMinSVN) bool` to `TaggedSVN`), and then either keep this method as-is or remove it we think that comparison between two MinSVN's never makes sense...

jraman567: > I'm wondering about the semantics of this. I was expecting to compare a "normal" SVN against a "minSVN", no? What does it mean to compare two "minSVN"s?

That's a good question, for Equality, I think they should exactly be the same.

But for comparison, I think the constraint on the reference doesn't matter. The minimum constraint only applies to the claim when comparing a MinSVN claim against a reference (ExactSVN or MinSVN).

For example, let's take the statement, "The minimum height to ride a roller coaster is 48 inches." There are two heights: the rider's and ruler's heights. When comparing, the minimum constraint only applies to the rider's height; the constraint on the ruler isn't something we care about for comparison (so it can be exact or minimal).

setrofim: >the constraint on the ruler isn't something we care about for comparison (so it can be exact or minimum).

In this example, a contraint on the ruler is something that doesn't really make sense, so we should not expose it as a method.

comid/svn.go
setrofim: Because the methods wern't really comparing  but testing for equality.  I realise the intent was to use this for verification, but wasn't requally comming across in the method. Also it's unecessarily contraining.  An assumption is made here that the only time one might wish to compare SVN values is in the context of verification; that might not be true. 

Note that you're not comparing evidence to reference values here. You're comparing one `TaggedMinSVN` to another. You're _assuming_ that given two variables of type `TaggedMinSVN`, `evidence` and `reference`, they will be used like `evidence.Equal(reference)`. However, there is nothing preventing the calling code from using them like `reference.Equal(evidence)`.  The notion of value vs reference is not encoded in the type.

I think it might be better to separate the operation being performed (comparison, equality test, etc) from higher level intent. So the method just tells you if the two values are equal or not (or how one compares relative to each other). What that means, in terms of validation should be left to the reference code.

jraman567:
@thomas-fossati @setrofim could you please confirm if the PR looks good? Thanks!

comid/svn.go
```diff
@@ -182,6 +199,38 @@ func (o TaggedMinSVN) Valid() error {
...
+// Compare compares TaggedMinSVN to a reference
+//
+// returns:
+// 0 if they are equal
+// 1 if SVN is newer than the reference
+// -1 if SVN is older than reference or if the method encounters an error
+func (o TaggedMinSVN) Compare(r any) int {
+	claim, err := convertToSVNUint64(o)
+	if err != nil {
+		log.Printf("TaggedMinSVN:Compare: Error: %v", err)
```
setrofim, line 211:
```suggestion
		log.Printf("TaggedMinSVN:Compare: Object Error: %v", err)
```

comid/svn.go
```diff
@@ -182,6 +199,38 @@ func (o TaggedMinSVN) Valid() error {
...
+		return -1
+	}
+
+	ref, err := convertToSVNUint64(r)
+	if err != nil {
+		log.Printf("TaggedMinSVN:Compare: Error: %v", err)
```
setrofim, line 217:
```suggestion
		log.Printf("TaggedMinSVN:Compare: Reference Error: %v", err)
```

comid/rawvalue.go
```diff
@@ -41,6 +43,22 @@ func (o RawValue) GetBytes() ([]byte, error) {
 	}
 }
 
+func (o *RawValue) Equal(r *RawValue) bool {
```
thomas-fossati, line 46: This may be a bit trickier than expected.

`Rawvalue` comparison also needs an optional mask to filter the "unimportant" bits (i.e., those that don't matter for comparison).  See [CoRIM](https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#section-5.1.4.1.4.2-5.5.1) for more details.

Besides, you'd compare it against a simple `[]byte` rather than another `RawValue`.

Therefore, the interface should look like:

```golang
func (o RawValue) Equal(other []byte, mask *[]byte) bool
```

This doesn't take into consideration the fact that a better way to model the `RawValue` type would be to include the optional mask alongside the value, which is not what we do ATM.  If we did that, then:

```golang
func (o RawValue) Equal(other []byte) bool
```

thomas-fossati (commented):
Looks good, thanks @jraman567 !

There's one point to discuss about what to do with `RawValue`.

jraman567:
@thomas-fossati and @setrofim, I have addressed your comments. Additionally, I made the following changes:

- Added two methods for each measurement type: `Equal` & `Compare`.
  - `Equal` compares an object with another object, it usually confirms if one object is an identical copy of another.
  - `Compare` method (note @setrofim , you wanted to name it as Check, but I'm going with Compare as it matches with the spec. Please confirm if that's OK) compares an object with a reference per the CoRIM spec
- The latest version adds Compare methods for SVN, Digests, RawValue & IntegrityRegisters. I appreciate your review of these types once again
- In some cases (Version, for example), both Equal and Compare do the same function. But I'm defining both for consistency.

Also, the lint error that CI reports (Implicit memory aliasing in for loop), is a false positive. It doesn't apply for golang versions 1.22 and newer.

setrofim:
> Compare method (note @setrofim , you wanted to name it as Check, but I'm going with Compare as it matches with the spec. Please confirm if that's OK) compares an object with a reference per the CoRIM spec

"Compare" as an operation has fairly well established semantics, which your methods do not implement. For example, in Go standard library:

https://pkg.go.dev/strings#Compare
https://pkg.go.dev/bytes#Compare
https://pkg.go.dev/cmp#Compare

In other libraries:

https://pkg.go.dev/github.com/open-policy-agent/opa/ast#Compare
https://pkg.go.dev/golang.org/x/mod/semver#Compare
https://pkg.go.dev/github.com/prometheus/prometheus/model/labels#Compare

And in other langunages, stretching all the way back to C (in the form of `strcmp` and `memcmp`).  Introducing a "compare" with different semantics is needlessly confusing and is begging to be misused.

Please also note that the spec specifically talks about "Comparing a condition ECT against the ACS", and not about comparing two values of those types in general. The defined operation is not a generic comparison, and so the method name should not imply that it is.  If you wish to stay closer to the spec, perhaps something like `CompareAgainstReference` or `CompareToCondition` would work.

Please rename.

jraman567:
Sounds reasonable. I'll change the name to `CompareAgainstReference`.

comid/digests.go
```diff
@@ -41,6 +43,54 @@ func (o Digests) Valid() error {
 	return nil
 }
 
+func (o *Digests) Equal(r *Digests) bool {
+	return reflect.DeepEqual(o, r)
```
setrofim, line 47: I don't think this is correct. The order of digests entries is semantically irrelevant, so I don't think to `Digests` instances that differ only in the order of their entries (but the entries, when reordered match).  

I think it should be something like this:
```go
func (o *Digests) Equal(r *Digests) bool {
	om := make(map[uint64][][]byte)
	for _, oe := range *o {
		vs, ok := om[oe.HashAlgID]
		if ok {
			om[oe.HashAlgID] = append(vs, oe.HashValue)
		} else {
			om[oe.HashAlgID] = [][]byte{oe.HashValue}
		}
	}

	outer:
	for _, re := range *r {
		ovs, ok := om[re.HashAlgID]
		if !ok {
			return false
		}

		for _, ov := range ovs {
			if bytes.Equal(ov, re.HashValue) {
				continue outer
			}
		}

		return false
	}

	return true
}
```

comid/digests.go
```diff
@@ -41,6 +43,54 @@ func (o Digests) Valid() error {
 	return nil
 }
 
+func (o *Digests) Equal(r *Digests) bool {
```
setrofim, line 46: Since we're not modifying `o`, should this method have an object, rather than pointer, receiver (ditto for all other methods)?

At one point, we had the convetion that for methods that do not modify the internal state should have an object receiver to indicate that.

comid/digests.go
setrofim: An argument can also be made for simply using the `CompareAgainstReference` implementation. If, semantically, `Digests` contains different hashes of the same object , so as long as the overlap matches, the underlying objects are equal, thefore the `Digests` can be said to be equal.

jraman567: You’re correct, that the order of entries don’t matter. Let me fix that. Thanks for bringing it up.

I don’t think reusing `CompareAgainstReference`, which does a partial match, confirms to the Equality semantics we’re going for here.

comid/digests.go
jraman567: My personal opinion is to avoid mixing receiver types; we should pick a pointer or value type and stay consistent for the whole object. If an object is immutable, it should have a value type; it can take a pointer type otherwise.

I’m obviously unaware of your convention. Please confirm if you want to stick to a single type for an object or change the receiver type depending on whether the method modifies the object. Thanks.

setrofim: I don't have a strong opinion on this. By using pointer receivers you avoid copying ponetially large values (here, `Digests` is already  a slice, which is tiny, so we don't win much in this case); by using a value receiver you do not risk modifying state from a method that is not intended to do so, and clearly communicate that (lack of) intent to the reader.

I don't see the advantage of avoiding switching receiver type between methods, depending on what the method needs. Consitently using pointer receiver in methods that don't need them just because _other_ methods modify state helpes neither readability nor usability.

However, I don't see this as a major issue. So if the others are ok with it, I'm fine with code going in as-is.

thomas-fossati: > I don't have a strong opinion on this. By using pointer receivers you avoid copying ponetially large values (here, `Digests` is already a slice, which is tiny, so we don't win much in this case); by using a value receiver you do not risk modifying state from a method that is not intended to do so, and clearly communicate that (lack of) intent to the reader.
> 
> I don't see the advantage of avoiding switching receiver type between methods, depending on what the method needs. Consitently using pointer receiver in methods that don't need them just because _other_ methods modify state helpes neither readability nor usability.
> 
> However, I don't see this as a major issue. So if the others are ok with it, I'm fine with code going in as-is.

I don't see a strong argument for changing the "good ole" convention.

jraman567: I switched the receiver types to the one @setrofim  proposed. However, the linter complained that the receiver for `FlagsMap` was too big, so I reverted it to a pointer.

I did the same for `Digests` & `IntegrityRegisters`, which have the potential to get big. All others follow the existing convention.

jraman567:
@thomas-fossati @setrofim Thanks for your review! I addressed all your comments, except for the receiver types for `FlagsMap`, `Digests` and `IntegrityRegisters`; I explained the reason above. Thanks!

comid/digests.go
thomas-fossati: hmm, `Digests` is a slice, which is always two `int`s and a pointer.

thomas-fossati: I am still pondering whether it's better (or worse) to have an inconsistent situation (both pointer & value receiver for "read-only" methods) or one that allows unexpected modifications of the underlying object (i.e., pointer receiver in all cases).

jraman567: Sorry, @thomas-fossati . I missed your comment from yesterday.

Due to how Golang passes slices around, could we even have read-only methods for slices since the methods could modify them partially (they have access to the underlying array)?

setrofim: If would be read-only with respect to the slice (i.e. you would not be able to, e.g. append elements), but not with respect to its contents.

"read-only"  also may not  be, strictly speaking, correct here. You're still able to modify the receiver variable, even with a value receiver. However, since its a copy, the modifications will not be visible outside the method.

thomas-fossati: > The other thing to consider is consistency with the rest of the code base. I
believe, until now, we have stuck with the convention that read-only methods
have value receivers. Linter complaints can be suppressed per-line using
//nolint as needed.

👍

thomas-fossati (approved):
Thanks for the changes (and the patience)!

comid/digests.go
```diff
@@ -41,6 +42,86 @@ func (o Digests) Valid() error {
 	return nil
 }
 
+// Equal confirms if the Digests object is the same as another
```
yogeshbdeshpande, line 45:
```suggestion
// Equal confirms if the Digests object is equal
```

comid/integrityregisters.go
```diff
@@ -136,3 +137,32 @@ func (i *IntegrityRegisters) UnmarshalJSON(data []byte) error {
...
+func (i IntegrityRegisters) CompareAgainstReference(r IntegrityRegisters) bool {
+	result := false
+
+	for refIndex, refDigests := range r.IndexMap {
+		claimDigests, ok := i.IndexMap[refIndex]
+		if !ok {
+			return false
+		}
+
+		ref := refDigests
```
yogeshbdeshpande, line 159: Is ref needed ?

comid/rawvalue.go
```diff
@@ -41,6 +44,36 @@ func (o RawValue) GetBytes() ([]byte, error) {
 	}
 }
 
+func (o RawValue) Equal(r RawValue) bool {
+	return reflect.DeepEqual(o, r)
+}
+
```
yogeshbdeshpande, line 50: Could you please add a header comment to this method and reference the comparison section!

comid/svn.go
```diff
@@ -182,6 +207,42 @@ func (o TaggedMinSVN) Valid() error {
...
+	ret, err := compare(o, r)
+	if err != nil {
+		log.Printf("TaggedMinSVN:Equal: %v", err)
+		return false
+	}
+
+	return ret == 0
+}
+
+// Compare helper function to compare two SVNs, object and reference
```
yogeshbdeshpande, line 220:
```suggestion
// compare is a helper function to compare two SVNs, object and reference
```

jraman567:
Thanks, @thomas-fossati & @setrofim for reviewing!

comid/digests.go
jraman567: Sorry, @yogeshbdeshpande . I didn't see your comments. Let me submit another PR with these changes.

comid/integrityregisters.go
jraman567: I got a linter error if I used refDigests directly. So, I created a copy of it.

comid/rawvalue.go
jraman567: OK, will do

comid/svn.go
jraman567: OK, will change it

jraman567:
Sorry, @yogeshbdeshpande . I didn't see your comment before submitting; I guess I opened the PR page when Thomas approved it, and I was in a meeting at that time. I merged it when the meeting got over, without refreshing the page.

PR #168, opened 2025-02-14
Updates the SignedCorim Sign method to include certificates

### LINKED ISSUE #159: Add signing key's certificate(s) as COSE headers
Add Signing Keys, Certificates to CoRIM

Refer to RFC 9360 for details.

Akhilesh-max:
Fix #159

### DIFF
--- a/corim/signedcorim.go
+++ b/corim/signedcorim.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"crypto"
 	"crypto/rand"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"strings"
@@ -24,9 +25,11 @@ var (
 // SignedCorim encodes a signed-corim message (i.e., a COSE Sign1 wrapped CoRIM)
 // with signature and verification methods
 type SignedCorim struct {
-	UnsignedCorim UnsignedCorim
-	Meta          Meta
-	message       *cose.Sign1Message
+	UnsignedCorim     UnsignedCorim
+	Meta              Meta
+	message           *cose.Sign1Message
+	signingCert       *x509.Certificate
+	intermediateCerts []*x509.Certificate
 }
 
 // NewSignedCorim instantiates an empty SignedCorim
@@ -126,9 +129,45 @@ func (o *SignedCorim) FromCOSE(buf []byte) error {
 	return nil
 }
 
+// AddSigningCert adds a DER-encoded X.509 certificate to be included in the
+// protected header of the COSE Sign1 message as the leaf certificate in X5Chain.
+func (o *SignedCorim) AddSigningCert(der []byte) error {
+	if der == nil {
+		return errors.New("nil signing cert")
+	}
+
+	cert, err := x509.ParseCertificate(der)
+	if err != nil {
+		return fmt.Errorf("invalid signing certificate: %w", err)
+	}
+
+	o.signingCert = cert
+	return nil
+}
+
+// AddIntermediateCerts adds DER-encoded X.509 certificates to be included in the protected
+// header of the COSE Sign1 message as part of the X5Chain.
+// The certificates must be concatenated with no intermediate padding, as per X.509 convention.
+func (o *SignedCorim) AddIntermediateCerts(der []byte) error {
+	if len(der) == 0 {
+		return errors.New("nil or empty intermediate certs")
+	}
+
+	certs, err := x509.ParseCertificates(der)
+	if err != nil {
+		return fmt.Errorf("invalid intermediate certificates: %w", err)
+	}
+
+	if len(certs) == 0 {
+		return errors.New("no certificates found in intermediate cert data")
+	}
+
+	o.intermediateCerts = certs
+	return nil
+}
+
 // Sign returns the serialized signed-corim, signed by the supplied cose Signer.
-// The target SignedCorim must have its UnsignedCorim field correctly
-// populated.
+// The target SignedCorim must have its UnsignedCorim field correctly populated.
 func (o *SignedCorim) Sign(signer cose.Signer) ([]byte, error) {
 	if signer == nil {
 		return nil, errors.New("nil signer")
@@ -160,6 +199,14 @@ func (o *SignedCorim) Sign(signer cose.Signer) ([]byte, error) {
 	o.message.Headers.Protected.SetAlgorithm(alg)
 	o.message.Headers.Protected[cose.HeaderLabelContentType] = ContentType
 	o.message.Headers.Protected[HeaderLabelCorimMeta] = metaCBOR
+
+	if o.signingCert != nil {
+		certChain := [][]byte{o.signingCert.Raw}
+		for _, cert := range o.intermediateCerts {
+			certChain = append(certChain, cert.Raw)
+		}
+		o.message.Headers.Protected[cose.HeaderLabelX5Chain] = certChain
+	}
 
 	err = o.message.Sign(rand.Reader, NoExternalData, signer)
 	if err != nil {
--- a/corim/signedcorim_test.go
+++ b/corim/signedcorim_test.go
@@ -4,7 +4,10 @@
 package corim
 
 import (
+	"bytes"
 	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


Akhilesh-max:
@thomas-fossati I am still working on the corresponding tests.

corim/signedcorim.go
```diff
@@ -127,13 +131,16 @@ func (o *SignedCorim) FromCOSE(buf []byte) error {
 }
 
 // Sign returns the serialized signed-corim, signed by the supplied cose Signer.
-// The target SignedCorim must have its UnsignedCorim field correctly
-// populated.
-func (o *SignedCorim) Sign(signer cose.Signer) ([]byte, error) {
+// The target SignedCorim must have its UnsignedCorim field correctly populated.
+func (o *SignedCorim) Sign(signer cose.Signer, leafCert, intermediateCert []byte) ([]byte, error) {
```
thomas-fossati, line 135: We cannot make any assumptions on the depth of the cert chain, i.e., the number of intermediate certs could be > 1.

corim/signedcorim.go
```diff
@@ -127,13 +131,16 @@ func (o *SignedCorim) FromCOSE(buf []byte) error {
...
 	if signer == nil {
 		return nil, errors.New("nil signer")
 	}
 
+	if leafCert == nil || intermediateCert == nil {
+		return nil, errors.New("nil certs")
+	}
```
thomas-fossati, line 142: We want to continue supporting the case of signed CoRIMs that rely on out-of-band means for communicating the verification key material.

Therefore, we shouldn't return an error here; continue as usual, just avoid adding the header.

corim/signedcorim.go
```diff
@@ -157,9 +164,19 @@ func (o *SignedCorim) Sign(signer cose.Signer) ([]byte, error) {
...
 	}
 
+	protectedHeaders := ProtectedHeader{
+		X5Chain: [][]byte{leafCert, intermediateCert},
+	}
+
+	protectedHeadersCBOR, err := cbor.Marshal(protectedHeaders)
+	if err != nil {
+		return nil, fmt.Errorf("failed CBOR encoding of protected headers: %w", err)
+	}
```
thomas-fossati, line 174: At this stage, this needs to remain an array - see https://www.rfc-editor.org/rfc/rfc9360.html#tab-1

The COSE library will take care of CBOR encoding.

corim/signedcorim.go
```diff
@@ -28,6 +29,9 @@ type SignedCorim struct {
 	Meta          Meta
 	message       *cose.Sign1Message
 }
+type ProtectedHeader struct {
+	X5Chain [][]byte `cbor:"33,keyasint,omitempty"`
+}
```
thomas-fossati, line 34: It looks like this isn't needed.

corim/signedcorim.go
```diff
@@ -157,9 +164,19 @@ func (o *SignedCorim) Sign(signer cose.Signer) ([]byte, error) {
...
+
+	protectedHeadersCBOR, err := cbor.Marshal(protectedHeaders)
+	if err != nil {
+		return nil, fmt.Errorf("failed CBOR encoding of protected headers: %w", err)
+	}
+
 	o.message.Headers.Protected.SetAlgorithm(alg)
 	o.message.Headers.Protected[cose.HeaderLabelContentType] = ContentType
 	o.message.Headers.Protected[HeaderLabelCorimMeta] = metaCBOR
+	o.message.Headers.Protected[33] = protectedHeadersCBOR
```
thomas-fossati, line 179: Rather than 33, you can use `cose.HeaderLabelX5Chain`

thomas-fossati (commented):
Thanks!

I have left a few comments inline.

From an API design perspective, instead of changing the `Sign`'s signature (apologies for the pun) I think it'd be better to add two new methods on the `SignedCorim` object to configure the end-entity cert and intermediate(s):

```golang
func (o *SignedCorim) AddSigningCert(der []byte) error { /* */ }
func (o *SignedCorim) AddIntemediateCerts(der []byte) error { /* */ }

```

corim/signedcorim.go
setrofim: Techinically, it could be 0 as well. Though unlikely, its possible the signing cert will be signed by the root CA cert.

Akhilesh-max: Got that. Update accordingly.

corim/signedcorim.go
Akhilesh-max: Updated,

corim/signedcorim.go
Akhilesh-max: Removed.

corim/signedcorim.go
Akhilesh-max: Updated.

setrofim:
There is an ordering dependency between methods that may lead to bugs. If you call  `AddSigningCert()`  after `AddIntemediateCerts()`, it will silently overwrite existing chain (`AddIntemediateCerts` has a check to ensure  that `AddSigningCert` was called beforehand, but the reverse is not true).

Either

- add a check to `AddSigningCert` to ensure that `cose.HeaderLabelX5Chain` is empty (it may then also be usefult to add a method to explicitly clear the chain), or
- keept track of them separately internally (e.g. as `o.signingCert` and `o.intermediateCerts`), and combine them and populate `HeaderLabelX5Chain` protected header inside `Sign()`.
- replace the two methods with `AddX5Chain()`, and leave it to the calling code to concat the signing and intermediate certs

The last option is probably the simplest to implment, as it religates most the othe responsibility to the caller. 

If you go for the ther option -- is it worth checking (probably during `Sign()`) that the intermediate cert(s) actually chain to the signing cert?

corim/signedcorim.go
```diff
@@ -126,9 +126,30 @@ func (o *SignedCorim) FromCOSE(buf []byte) error {
 	return nil
 }
 
+func (o *SignedCorim) AddSigningCert(der []byte) error {
+	if der == nil {
+		return errors.New("nil signing cert")
+	}
+	o.message.Headers.Protected[cose.HeaderLabelX5Chain] = [][]byte{der}
```
thomas-fossati, line 133: We shouldn't directly fiddle with the COSE internal structure here.

Instead, create two new attributes in the `SignedCorim` object: `signingCert` and `intermediateCerts` and set those from here.

Then, in `Sign`, populate the COSE headers as needed.

This allows us more flexibility in case we decide to change the internal arrangement from x5chain to x5chain+x5t, or x5bag, etc.

thomas-fossati:
> * add a check to `AddSigningCert` to ensure that `cose.HeaderLabelX5Chain` is empty (it may then also be usefult to add a method to explicitly clear the chain), or

I think we should not manipulate the internal COSE arrangement from here.

> * keept track of them separately internally (e.g. as `o.signingCert` and `o.intermediateCerts`), and combine them and populate `HeaderLabelX5Chain` protected header inside `Sign()`.

Yes!

> * replace the two methods with `AddX5Chain()`, and leave it to the calling code to concat the signing and intermediate certs

I'd rather keep the separation in case we decide later on to have different split of the material (see https://github.com/veraison/corim/pull/168#discussion_r1975029174).

Akhilesh-max:
@thomas-fossati Made the suggested updates. PTAL

Thanks!

corim/signedcorim.go
```diff
@@ -24,9 +24,11 @@ var (
...
 // with signature and verification methods
 type SignedCorim struct {
-	UnsignedCorim UnsignedCorim
-	Meta          Meta
-	message       *cose.Sign1Message
+	UnsignedCorim     UnsignedCorim
+	Meta              Meta
+	message           *cose.Sign1Message
+	signingCert       []byte
+	intermediateCerts [][]byte
```
thomas-fossati, line 31: We can use the convention used by [Go's crypto library](https://pkg.go.dev/crypto/x509#ParseCertificates) in similar situations, where the chain is supplied as a single byte buffer where _"certificates must be concatenated with no intermediate padding."_

corim/signedcorim.go
```diff
@@ -126,9 +128,25 @@ func (o *SignedCorim) FromCOSE(buf []byte) error {
 	return nil
 }
 
+func (o *SignedCorim) AddSigningCert(der []byte) error {
```
thomas-fossati, line 131: Public funcs should be documented and start with a godoc comment like: _"AddSigningCert adds ..."_

See also https://go.dev/doc/comment

corim/signedcorim.go
```diff
@@ -126,9 +128,25 @@ func (o *SignedCorim) FromCOSE(buf []byte) error {
...
+	if der == nil {
+		return errors.New("nil signing cert")
+	}
+	o.signingCert = der
```
thomas-fossati, line 135: We should validate the supplied cert using [the crypto library](https://pkg.go.dev/crypto/x509#ParseCertificate).

corim/signedcorim.go
thomas-fossati, line 31: These can be [parsed (and validated) x509 certs](https://pkg.go.dev/crypto/x509#Certificate).

corim/signedcorim.go
```diff
@@ -161,6 +179,14 @@ func (o *SignedCorim) Sign(signer cose.Signer) ([]byte, error) {
 	o.message.Headers.Protected[cose.HeaderLabelContentType] = ContentType
 	o.message.Headers.Protected[HeaderLabelCorimMeta] = metaCBOR
 
+	if o.signingCert != nil {
+		certChain := [][]byte{o.signingCert}
```
thomas-fossati, line 183:
```suggestion
		certChain := [][]byte{o.signingCert.Raw}
```

Akhilesh-max: Updated.

corim/signedcorim.go
Akhilesh-max: Done.

corim/signedcorim.go
Akhilesh-max: Done.

corim/signedcorim.go
Akhilesh-max: Added.

corim/signedcorim.go
Akhilesh-max: Updated accordingly.

corim/signedcorim.go
```diff
@@ -126,9 +129,45 @@ func (o *SignedCorim) FromCOSE(buf []byte) error {
...
+		return fmt.Errorf("invalid signing certificate: %w", err)
+	}
+
+	o.signingCert = cert
+	return nil
+}
+
+// AddIntermediateCerts adds DER-encoded X.509 certificates to be included in the protected
+// header of the COSE Sign1 message as part of the X5Chain.
+// The certificates must be concatenated with no intermediate padding, as per X.509 convention.
```
thomas-fossati, line 150: AFAIK it's a go crypto/x509 convention rather than a general convention around X.509 chain handling.

```suggestion
// The certificates must be concatenated with no intermediate padding.
```

corim/signedcorim.go
```diff
@@ -161,6 +200,16 @@ func (o *SignedCorim) Sign(signer cose.Signer) ([]byte, error) {
 	o.message.Headers.Protected[cose.HeaderLabelContentType] = ContentType
 	o.message.Headers.Protected[HeaderLabelCorimMeta] = metaCBOR
 
+	if o.signingCert != nil {
+		certChain := [][]byte{o.signingCert.Raw}
+		if len(o.intermediateCerts) > 0 {
```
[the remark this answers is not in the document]
Akhilesh-max: Is. 
Updated accordingly.

thomas-fossati:
Looks great, thanks for all the changes.

You need to add a few unit tests to make the CI green.

* AddSigningCert (positive and negative tests)
* AddIntermediateCerts (ditto)

It shouldn’t be hard to move the needle from 84.2% to the 84.4% threshold.

And you can use the certs you already created under `misc` as test vectors.

Akhilesh-max:
@thomas-fossati I've add the tests. PTAL.
Thanks!

corim/signedcorim_test.go
```diff
@@ -530,3 +532,71 @@ func TestSignedCorim_extensions(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := NewSignedCorim()
+			err := s.AddSigningCert(tt.certDer)
+
+			if tt.wantErr {
+				assert.Error(t, err)
+				assert.Nil(t, s.signingCert)
+			} else {
```
thomas-fossati, line 562: and then

```suggestion
				assert.Nil(t, s.signingCert)
				assert.EqualError(t, err, tt.errMsg)
			} else {
```

corim/signedcorim_test.go
```diff
@@ -530,3 +532,71 @@ func TestSignedCorim_extensions(t *testing.T) {
...
+				assert.NotNil(t, s.signingCert)
+			}
+		})
+	}
+}
+
+func TestSignedCorim_AddIntermediateCerts(t *testing.T) {
+	certPath := filepath.Join("..", "misc", "intermediateCA.der")
+	validCert, err := os.ReadFile(certPath)
+	require.NoError(t, err, "Failed to read test certificate")
```
thomas-fossati, line 573: This only tests 1-element chains, but we want to make sure that the concatenation of two or more certs also works.

I suggest adding a `concatFiles` function like

```golang
func concatFiles(files ...string) ([]byte, error) {
	var buf bytes.Buffer

	for _, file := range files {
		b, err := ioutil.ReadFile(file)
		if err != nil {
			return nil, err
		}
		buf.Write(b)
	}

	return buf.Bytes(), nil
}
```

and then call it here:
```suggestion
	validCerts, err := concatFiles("../misc/intermediateCA.der", "../misc/rootCA.der")
	require.NoError(t, err, "Failed to read test certificates")
```

thomas-fossati (commented):
Excellent start.  Find a couple of improvement suggestions inline.

corim/signedcorim_test.go
```diff
@@ -530,3 +532,71 @@ func TestSignedCorim_extensions(t *testing.T) {
...
+
+	tests := []struct {
+		name    string
+		certDer []byte
+		wantErr bool
```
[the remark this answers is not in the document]
Akhilesh-max: Done.

corim/signedcorim_test.go
Akhilesh-max: Added another test for multiple intermediate certs.

corim/signedcorim_test.go
Akhilesh-max: Updated accordingly.

corim/signedcorim_test.go
```diff
@@ -530,3 +532,71 @@ func TestSignedCorim_extensions(t *testing.T) {
...
+func TestSignedCorim_AddSigningCert(t *testing.T) {
+	certPath := filepath.Join("..", "misc", "endEntity.der")
+	validCert, err := os.ReadFile(certPath)
+	require.NoError(t, err, "Failed to read test certificate.")
+
+	tests := []struct {
+		name    string
+		certDer []byte
+		wantErr bool
+	}{
```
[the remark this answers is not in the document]
Akhilesh-max: Updates accordinlgy.

Akhilesh-max:
@thomas-fossati Thanks for the reviews and guidance so far. 
I have made the changes as per the recent review. 
Please take a look when get a chance.

thomas-fossati (approved):
Awesome! Thanks for all the good work.

I reckon we can and declare victory and merge this :-)

PR #170, opened 2025-02-27
fix(digests/rawvalue/svn): method header comments

jraman567:
fix/add method header comments

### DIFF
--- a/comid/svn.go
+++ b/comid/svn.go
@@ -217,7 +217,7 @@ func (o TaggedMinSVN) Equal(r TaggedMinSVN) bool {
 	return ret == 0
 }
 
-// Compare helper function to compare two SVNs, object and reference
+// Compare is a helper function to compare two SVNs, object and reference
 //
 // returns:
 // 0 if they are equal; no error
--- a/comid/svn.go
+++ b/comid/svn.go
@@ -217,7 +217,7 @@ func (o TaggedMinSVN) Equal(r TaggedMinSVN) bool {
 	return ret == 0
 }
 
-// Compare is a helper function to compare two SVNs, object and reference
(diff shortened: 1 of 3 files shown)


jraman567:
@yogeshbdeshpande I addressed your comments from [PR 167](https://github.com/veraison/corim/pull/167). I didn't add header comments to other Equal/Compare methods (for example, in svn.go) as they seemed self-explanatory. Do you prefer I add header comments to them also?

Concerning your comments in integrityregisters.go:

> Is ref needed ?

I get a linter error if I don't use a local variable. I get it only on GitHub CI, not on my local machine.

This error ("G601: Implicit memory aliasing in for loop. (gosec)") [shouldn't occur with version 1.22 or newer](https://tip.golang.org/doc/go1.22), but I can't confirm. I can turn off this linter test for this specific line, assuming we don't legitimately have this error. Alternatively, we could update the gosec/golang version used by GitHub CI.

comid/svn.go
```diff
@@ -217,7 +217,7 @@ func (o TaggedMinSVN) Equal(r TaggedMinSVN) bool {
 	return ret == 0
 }
 
-// Compare helper function to compare two SVNs, object and reference
+// Compare is a helper function to compare two SVNs, object and reference
```
thomas-fossati, line 220: It's a [Godoc convention](https://tip.golang.org/doc/comment) to start the comment on a function with the name of the function.

```suggestion
// compare is a helper function to compare two SVNs, object and reference
```

thomas-fossati (approved):
LGTM, thanks!

(Please apply the inlined suggestion before merging.)

setrofim (approved):
LGTM

PR #176, opened 2025-03-06
fix(corim): handle silly RFC9360 microoptimisation

thomas-fossati:
x5chain is of type `COSE_X509`, which is defined as:
    
    COSE_X509 = bstr / [ 2*certs: bstr ]
    
If there is only one element, it should be encoded as a simple `bstr` instead of being represented as an array.


setrofim (approved):
*sigh* of course. They could have had it as a `bstr` with all certs concatenated, _or_ an `[1* bstr ]`; instead, they went with option C...

PR #183, opened 2025-04-26
rawint: add a raw int concrete type

jraman567:
Add a raw int concrete type as mentioned in the IETF CoRIM spec: https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#name-raw-int

### DIFF
--- a/comid/rawint.go
+++ b/comid/rawint.go
@@ -14,6 +14,11 @@ import (
 	"github.com/veraison/corim/extensions"
 )
 
+var (
+	ErrRawIntUnsetRawInt = errors.New("RawInt value unset")
+	ErrRawIntEmpryInput  = errors.New("Empry input")
+)
+
 // RawInt describes an integer value that can be compared with linear order in
 // the target environment. It follows the type-choice pattern  widely
 // used in CoRIM and implements the extensions.ITypeChoiceValue interface
@@ -47,7 +52,7 @@ func (o RawInt) Type() string {
 // Valid checks if the RawInt is valid
 func (o RawInt) Valid() error {
 	if !o.IsSet() {
-		return errors.New("RawInt value unset")
+		return ErrRawIntUnsetRawInt
 	}
 
 	return o.Value.Valid()
@@ -108,7 +113,7 @@ func (o RawInt) MarshalCBOR() ([]byte, error) {
 // UnmarshalCBOR de-serializes input CBOR into RawInt
 func (o *RawInt) UnmarshalCBOR(data []byte) error {
 	if len(data) == 0 {
-		return errors.New("empty input")
+		return ErrRawIntEmpryInput
 	}
 
 	majorType := (data[0] & 0xe0) >> 5
@@ -126,7 +131,7 @@ func (o *RawInt) UnmarshalCBOR(data []byte) error {
 		}
 		o.Value = rawIntRange
 	default:
-		return fmt.Errorf("RawInt: Error: unknown major type: %d", majorType)
+		return fmt.Errorf("RawInt: unknown major type: %d", majorType)
 	}
 
 	return nil
@@ -172,12 +177,16 @@ func (o RawIntInteger) Type() string {
 	return RawIntIntegerType
 }
 
-// CompareAgainstRefInteger compares RawIntInteger against another RawIntInteger reference value
+// CompareAgainstRefInteger compares RawIntInteger against another RawIntInteger
+// reference value. The receiver is the claim and the input parameter is the
+// reference value. Returns true if the claim is equal to the reference; false otherwise.
 func (o RawIntInteger) CompareAgainstRefInteger(ref RawIntInteger) bool {
 	return o == ref
 }
 
-// CompareAgainstRefRange compares RawIntInteger against a TaggedRawIntRange
+// CompareAgainstRefRange compares RawIntInteger against a TaggedRawIntRange.
+// The receiver is the claim and the input parameter is the reference value.
+// Returns true if the claim is within the reference range; false otherwise.
 func (o RawIntInteger) CompareAgainstRefRange(ref TaggedRawIntRange) bool {
 	obj, err := convertRawIntIntegerToInt64(o)
 	if err != nil {
@@ -273,7 +282,9 @@ func (o TaggedRawIntRange) Type() string {
 }
 
 // CompareAgainstRefInteger compares TaggedRawIntRange against a given
-// RawIntInteger reference
+// RawIntInteger reference. The receiver is the claim and the input
+// parameter is the reference value. Returns true if and only if the
+// claim is equal to both reference min & max values; false otherwise.
 func (o TaggedRawIntRange) CompareAgainstRefInteger(ref RawIntInteger) bool {
 	refVal, err := convertRawIntIntegerToInt64(ref)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


comid/rawint.go
```diff
@@ -0,0 +1,330 @@
...
+func (o RawIntInteger) Type() string {
+	return RawIntIntegerType
+}
+
+// CompareAgainstRefInteger compares RawIntInteger against another RawIntInteger reference value
+func (o RawIntInteger) CompareAgainstRefInteger(ref RawIntInteger) bool {
+	return o == ref
+}
+
+// CompareAgainstRefRange compares RawIntInteger against a TaggedRawIntRange
```
deeglaze, line 179: This should describe the comparison rule. An Int is equal to an IntRange iff the int range min and max are both the same as the int.

comid/rawint.go
```diff
@@ -0,0 +1,330 @@
...
+	return fmt.Sprintf("%s:%s", rangeMin, rangeMax)
+}
+
+// Type returns the name/type of TaggedRawIntRange
+func (o TaggedRawIntRange) Type() string {
+	return TaggedRawIntRangeType
+}
+
+// CompareAgainstRefInteger compares TaggedRawIntRange against a given
+// RawIntInteger reference
```
deeglaze, line 275: This should describe the comparison rule. An int range is equal to an int iff the min and max are both equal to the int.

comid/rawint.go
```diff
@@ -0,0 +1,330 @@
...
+
+	if o.Min != nil && refVal == *o.Min && o.Max != nil && refVal == *o.Max {
+		return true
+	}
+
+	return false
+}
+
+// CompareAgainstRefRange compares TaggedRawIntRange against another
+// TaggedRawIntRange reference
```
deeglaze, line 291: This should describe the comparison rule. This one is the trickier one where the argument order matters. The receiver is the claim and the argument is the condition. The condition range matches if it subsumes the claim range.

comid/rawint.go
```diff
@@ -0,0 +1,330 @@
...
+
+// NewRawIntRangeType returns a new TaggedRawIntRange from given value
+func NewRawIntRangeType(val any) (*RawInt, error) {
+	ret, err := NewRawIntRange(val)
+	if err != nil {
+		return nil, err
+	}
+
+	return &RawInt{ret}, nil
+}
```
deeglaze, line 322: I don't understand why an int range only takes a single value and returns a non-range.

deeglaze (changes requested):

comid/rawint.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+		o.Value = rawIntInteger
+	case 6:
+		rawIntRange := new(TaggedRawIntRange)
+		if err := dm.Unmarshal(data, rawIntRange); err != nil {
+			return err
+		}
+		o.Value = rawIntRange
+	default:
+		return fmt.Errorf("RawInt: Error: unknown major type: %d", majorType)
```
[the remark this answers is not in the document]
jraman567: @deeglaze raw-int-type-choice Going by the [CoRIM spec](https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#name-raw-int), `raw-int-type-choice` can either be `int-range` (major types 0 & 1) or `tagged-int-range` (major type 6). Any other major type isn't applicable to raw-int.

So it seems as if any other major type is an error. Am I missing anything?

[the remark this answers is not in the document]
deeglaze: I'm just saying that the error reporting string doesn't also need to say "Error:" since it's redundant.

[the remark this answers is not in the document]
jraman567: Got it, thanks

comid/rawint.go
```diff
@@ -0,0 +1,330 @@
...
+	"github.com/veraison/corim/encoding"
+	"github.com/veraison/corim/extensions"
+)
+
+// RawInt describes an integer value that can be compared with linear order in
+// the target environment. It follows the type-choice pattern  widely
+// used in CoRIM and implements the extensions.ITypeChoiceValue interface
+// https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#name-raw-int
+type RawInt struct {
+	Value extensions.ITypeChoiceValue
```
yogeshbdeshpande, line 21: The extensions framework is for types that is for `extensible-type-choice` as per 
https://ietf-rats-wg.github.io/draft-ietf-rats-corim/draft-ietf-rats-corim.html#name-typographical-conventions-f

However, raw-int is a closed type choice, any reason why it is made extensible ?

yogeshbdeshpande (changes requested):
please see, my comment

comid/rawint.go
jraman567: `extensions.ITypeChoiceValue` appears to be the interface that all concrete types in CoRIM implement. Is there a different interface that rawint should implement?

Please see the following comment for extensions.ITypeChoiceValue:
```golang
// ITypeChoiceValue is the interface that is implemented by all concrete type
// choice value types. Specific type choices define their own value interfaces
// that embed this one (and possibly include additional methods).
```

For example, closed types like SVN implements this interface also:
```golang
type SVN struct {
	Value ISVNValue
}

type ISVNValue interface {
	extensions.ITypeChoiceValue
}
```

comid/rawint.go
jraman567: It's a factory function used by `NewRawInt`. The function has a fixed prototype, like below:
`type IRawIntFactory = func(val any) (*RawInt, error)`

That's why it take only one input and returns RawInt (which could either be an Integer or Range based on the function).

comid/rawint.go
```diff
@@ -0,0 +1,344 @@
...
+	"log"
+	"strconv"
+
+	"github.com/veraison/corim/encoding"
+	"github.com/veraison/corim/extensions"
+)
+
+var (
+	ErrRawIntUnsetRawInt = errors.New("RawInt value unset")
+	ErrRawIntEmpryInput  = errors.New("Empry input")
```
deeglaze, line 19: Exported symbols must be documented (comment directly above starting with the symbol name)

comid/rawint.go
```diff
@@ -15,8 +15,11 @@ import (
 )
 
 var (
+	// ErrRawIntUnsetRawInt is the error returned when a RawInt value isn't set
 	ErrRawIntUnsetRawInt = errors.New("RawInt value unset")
-	ErrRawIntEmpryInput  = errors.New("Empry input")
+
+	// ErrRawIntEmpryInput is the error returned when the input to a function is empty
+	ErrRawIntEmpryInput = errors.New("Empry input")
```
yogeshbdeshpande, line 22:
```suggestion
	ErrRawIntEmptyInput = errors.New("Empty input")
```

comid/rawint.go
```diff
@@ -15,8 +15,11 @@ import (
 )
 
 var (
+	// ErrRawIntUnsetRawInt is the error returned when a RawInt value isn't set
 	ErrRawIntUnsetRawInt = errors.New("RawInt value unset")
-	ErrRawIntEmpryInput  = errors.New("Empry input")
+
+	// ErrRawIntEmpryInput is the error returned when the input to a function is empty
```
yogeshbdeshpande, line 21:
```suggestion
	// ErrRawIntEmptyInput is the error returned when the input to a function is empty
```

yogeshbdeshpande (commented):
please see some more comments

deeglaze (approved):

jraman567:
@yogeshbdeshpande Hi Yogesh, just checking in on the PR. Let me know if there’s anything else I can address to help finalize the review!

jraman567:
> LGTM! Thanks for the PR!

Thank you for confirming!

jraman567:
Thank you for reviewing the PR, @yogeshbdeshpande and @deeglaze!

Issue #131, opened 2024-10-22
FromCOSE payload type not tagged unsigned-corim


deeglaze:
The CoRIM specification requires a tagged-unsigned-corim in the payload according to the CDDL. I don't think it should be required sign the content-type in the protected headers is enough context for the signature.

Either a spec flaw or an implementation bug...

deeglaze:
Filed https://github.com/ietf-rats-wg/draft-ietf-rats-corim/pull/332 as well.

deeglaze:
Dropping this to simplify.

deeglaze:
Reopening since I haven't reverted PR#332 in the spec.

setrofim:
Addressed by https://github.com/veraison/corim/pull/187

PR #187, opened 2025-06-12
fix!: unsigned CoRIMs always tagged

setrofim:
Update ToCBOR/FromCBOR for UnsignedCorim to add/expect tag 501.

Current spec requires that unsigned CoRIMs are always tagged with tag 501, both, when appearing as stand-alone CBOR documents, and when embedded as payloads of signed CoRIMs.

Note: There was already TaggedUnsignedCorim type registered under tagged 501, however it was not being used and its marshalling and didn't work correctly due to the custom To/FromCBOR in UnsignedCorim used in lieu of the standard marshalling mechanism. Rather than (a) fixing  its serialization, (b) forwarding all UnsignedCorim methods, and (c) replacing UnsignedCorim usaged with TaggedUnsignedCorim all over the place,  TaggedUnsignedCorim was simply removed, and tagging was added to CBOR serialization of UnsignedCorim, as it's not really valid for it to appear untagged, and this minimizes API impact (and is also consistent with SignedCorim that serializes tagged).

BREAKING CHANGE: serialized unsigned CoRIMs now start with tag 501. (This is an ABI-only break; all APIs remain the same).

This addresses https://github.com/veraison/corim/issues/131


deeglaze (approved):

thomas-fossati (approved):
Looks good, thanks!

corim/profiles.go
```diff
@@ -74,11 +76,15 @@ func UnmarshalSignedCorimFromCBOR(buf []byte) (*SignedCorim, error) {
 // the data, they will be registered with the UnsignedCorim before it is
 // unmarshaled.
 func UnmarshalUnsignedCorimFromCBOR(buf []byte) (*UnsignedCorim, error) {
+	if !bytes.Equal(buf[:3], UnsignedCorimTag) {
```
yogeshbdeshpande, line 79: Should the comment on Line 74 not clarify that the CBOR payload is now expected to be tagged?

// // UnmarshalUnsignedCorimFromCBOR unmarshals a tagged UnsignedCorim from provided
// CBOR data.

corim/testcases/src/example-corim.yaml
```diff
@@ -6,73 +6,75 @@
...
-- encodedCBOR:
-    tag: 506
-    value:
-      0: en-GB
-      1:
-        0: !!binary |-
-          Q7vjfy5hSzOu01PP8UKLFg==
-      2:
-      - 0: ACME Ltd.
+tag: 501
```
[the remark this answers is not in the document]
setrofim: "tagged" is implied. It's not valid for a CoRIM not to be tagged, so the distinction does not needs to be made ( just like we don't say "tagged signed corim".

corim/profiles.go
setrofim: "tagged" is implied. It's not valid for an unsigned corim not to be tagged.

PR #189, opened 2025-06-16
fix: handle short buffers inside UnsignedCorim.FromCBOR()

setrofim:
UnsignedCorim.FromCBOR() checks for the unsigned  corim tag by taking a slice of the first three bytes of the input. This will panic if the input is shorter than three bytes, so check for that first.


thomas-fossati (approved):

yogeshbdeshpande (approved):
LGTM!

PR #204, opened 2025-06-19
fix: validate inside UnsignedCorim.ToCBOR()

setrofim:
Validate before serializing inside UnsignedCorim.ToCBOR() to be consistent with the semantics of ToCBOR() in other types.


thomas-fossati (approved):
LGTM, thanks!

PR #210, opened 2025-07-11
feat(coserv): implement stateful selectors

thomas-fossati:


paulhowardarm (approved):
This all looks as I would expect. Great stuff.

PR #235, opened 2025-10-23
feat(comid): add string Mkey support

setrofim:
Since rev. 6, CoRIM spec expanded $measured-element-type-choice with tstr variant. Thus updates the corresponding Mkey type to support the new StringMkey variant.


yogeshbdeshpande:
Thanks, I will review it today!

Issue #211, opened 2025-07-11
Interesting.  CoRIM-07 Reevaluate IndefLength: cbor.IndefLengthForbidden, inside cbor.go.


yogeshbdeshpande:
Potential incorrect restriction on cbor.IndefLengthForbidden Flag in CoRIM Repository

_Originally posted by @thomas-fossati in https://github.com/veraison/corim/pull/205#discussion_r2172082520_

PR #239, opened 2025-11-06
fix: allow indefinite length when decoding

setrofim:
The CoRIM specification does not specify any restrictions on the CBOR encoding, so indefinite length structures are valid.

Since CoSERV _does_ require deterministic (and therefore, definite length) encoding, keep the existing restriction on encoding.

This is anlternative to https://github.com/veraison/corim/pull/222 that does not touch the encoding path.

Addresses https://github.com/veraison/corim/issues/211.


yogeshbdeshpande (approved):
LGTM!

PR #249, opened 2025-12-12
Unknown extension caching

setrofim:
This pull implements the caching of unknown extension values when unmarshalling CoRIMs. Cached values are then incorporated when re-marshalling. This ensures that a CoRIM is preserved in its entirety across re-marshalling, even when containing extension values that do not correspond to fields in registered extension structs inside the unmarshal target.

This is necessary to support future migration of Veraison services to use corim-store, which will require lossless handling of CoRIMs without awareness of scheme-specific extensions.

Additionally, this ensures CDE compliance when marshalling to CBOR.

### DIFF
--- a/extensions/extensions_test.go
+++ b/extensions/extensions_test.go
@@ -1,4 +1,4 @@
-// Copyright 2023-2024 Contributors to the Veraison project.
+// Copyright 2023-2025 Contributors to the Veraison project.
 // SPDX-License-Identifier: Apache-2.0
 package extensions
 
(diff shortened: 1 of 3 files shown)


encoding/field_cache_test.go
```diff
@@ -0,0 +1,179 @@
...
+		0x20,             // key: -1
+		0x63,             // value: tstr(3)
+		0x62, 0x61, 0x72, // "bar"
+	}
+
+	em := mustInitEncMode()
+	dm := mustInitDecMode()
+
+	// First, a sanity test to make sure that embedded data
+	// is presevered when there is a concrete struct to
```
yogeshbdeshpande:
```suggestion
	// is preserved when there is a concrete struct to
```

extensions/extensions_test.go
```diff
@@ -168,3 +169,114 @@ func Test_Extensions_Values(t *testing.T) {
...
+	_, err = encoding.SerializeStructToCBOR(em, &entity)
+	assert.NoError(t, err)
+
+	entity = Entity{}
+	entity.Register(&TestExtensions{})
+	err = encoding.PopulateStructFromJSON(data, &entity)
+	assert.NoError(t, err)
+
+	// If extensions were registered before unmarshalling, the value gets
+	// populated directily into the registred struct, bypassing the cache.
```
yogeshbdeshpande:
```suggestion
	// populated directly into the registered struct, bypassing the cache.
```

PR #264, opened 2026-04-30
fix(comid): handle unset version scheme

setrofim:
The scheme part of a version-map is optional and so it should be possible to encode a Version without calling SetScheme(), however attempting to do so resulted in an error due to swid.VersionScheme being unable to deal with a nil inner value.

Change the type of the Scheme field to be a point to, rather than a swid.VersionScheme, and mark it as omitempty.


thomas-fossati (approved):
Of course!  Thanks!

PR #268, opened 2026-05-20
Align with rev. 10 of the spec

setrofim:
This adds missing triple types and brings the implementation in alignment with rev. 10 of the spec. This is verified by adding all `corim-*.cddl` and `comid-*.cddl` examples from the spec repo and ensuring that they unmarshal without error.


corim/oneormore.go
thomas-fossati, line 1: Note for self: this would come in handy in the x5chain header handling in signed corim.

comid/domain_membership_triple.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Add a triple to the DomainMembershipTriples.
+func (o *DomainMembershipTriples) Add(triple DomainMembershipTriple) *DomainMembershipTriples {
+	*o = append(*o, triple)
+	return o
+}
+
+// Valid retursn an error if DomainMembershipTriples is empty or conatains
+// invalid elements.
+func (o DomainMembershipTriples) Valid() error {
```
thomas-fossati, line 72: Another validation criterion is that the graph described by the triples is acyclic.  We have a similar check already in place for the domain dependency triples.

comid/rawvalue.go
thomas-fossati, line 1: note to self: if I'm not mistaken, this was the missing piece needed to finalise the alignment of the CCA profile implementation with the specification.

comid/digest.go
thomas-fossati, line 1: note to self: veraison/eat’s measured component implementation (https://github.com/veraison/eat/blob/main/measured_component.go) can be updated accordingly.

thomas-fossati (approved):
Awesome work, thanks!

I have left a comment about the validation of membership triples, but it is non-blocking (i.e., we can address that separately)

comid/domain_membership_triple.go
setrofim: According to the spec, acyclicity is a requirement only for domain dependency triples; it is not mentioned for domain membership.

thomas-fossati: All right, that's probably a bug in the specification.
Logically, it makes no sense for the graph describing a domain's aggregate to be cyclical: it’d mean that an attester contains itself :-)

setrofim: Agreed, however, if the usage described in the spec does not rely on acyclicity, there is no reason for it to be explicitly stated. For domain dependencies, processing in section 9 requires acyclicity, hence the explicit constraint. (At least, that was my interpretation of why the spec is the way it is).

thomas-fossati: I believe the current spec's asymmetrical behaviour is confusing.
Even for domain dependencies, we could allow the verifier to determine that a cycle is about to be entered, rather than making this a precondition of the processing step...

setrofim: Sure, but the point is that the verifier needs there to not be a cycle, since dependencies impact the order in which ECTs ought to be processed (you can only trust a domain if you can trust its trustees). There is no such requirement on domain membership handling, which only tests for presence of ECTs and does not rely on absolute ordering.

If there is a cycle in dependencies, verification cannot happen because a trust chain cannot be established. If there is a cycle in membership, verification is not impacted).

thomas-fossati: Not really, in both cases successful appraisal depends on the policy that consumes the produced ECT.  If the CoRIM processor cannot compute the membership/dependency graph (in this case because of a cycle in the contributing triples), appraisal may fail hard on the absence of a needed construct.

setrofim: Computing the graph is not the problem -- this can always be done. The issue is with traversing the graph afterwards. 

In case of dependencies, you have "A can only be trusted if B is already trusted; B can only be trusted if C is already trusted; C can only trusted if A is already trusted". This cannot be resolved.

In case of membership there the condition is not directional. If you have "A is a member implies* ("implies", and not "conditioned on") B is a member; B is a member implies C is a member; C is a member implies A is a member", can simply be resolved by adding all three. Here the cycle translates into "all or nothing", and does not create a contradiction since membership, unlike trust, does not need to be "rooted". 

A dependency triple introduces a constraint "Trusting A is contingent on trusting its trustee B".

A membership triple introduces a relationship "B is a member of A", but it does not place any constraints on A.

thomas-fossati: From the domain graph perspective, it makes no sense (logically and physically) for a member to be contained in itself.  In that sense, the membership graph cannot be (meaningfully) computed if there is a cycle.

setrofim: But this has no bearing on intended usage vis. verification, hence acyclicity does not need to be an explicit constraint.

Also -- maybe...?

Logically: this entirely possible in the abstract. `A ⊆ B`, `B ⊆ C`, `C ⊆ A` just resolves to `A = B = C` (i.e. "all or nothing"). There is no contradiction. If you have multiple domains whose membership overlaps perfectly, it is entirely valid (if questionably useful) to describe their membership in terms of each other.

Physically: I'm not sure? Both "environment" and "domain" are intentionally defined sufficiently abstractly that I would uncomfortable ruling it out. In the "common" case where the domain hierarchy maps onto sub-component hierarchy for a system, a cycle certainly does not make sense. But I don't believe the environments/domains necessarily need to be mapped that way?

I guess, what I'm saying is that there is a difference between "I can't think of a case where this makes sense" vs "this never makes sense", and think the situation is in the former category, so there no point in intentionally enforcing the constraint if it is not required for intended usage.

thomas-fossati: Both environments and domains, despite their apparent abstraction, represent a precise logical - if not physical - reality: a network of RoTs connected by secure communication channels.
To think of a RoT as a proper subset of itself bears no relation to reality, and modelling it simply makes no sense.

setrofim: Agreed, but does membership imply specifically a proper/strict subset? Can you not have a composite attester with exactly one constituent, so that there is a single RoT which is in fact a (non-proper) subset of itself?

thomas-fossati: No. If it is a composite device, there must be one lead and at least one sub-attester. If it is a layered attester, there must be at least two layers.
In both cases, a composite attester requires at least two distinct components.

setrofim: OK, if it is forbidden for the lead and the sub-attester to be the same, then fair enough.

setrofim: I've updated the implementation to check for cycles in the same way as dependency triples.

thomas-fossati: Thank you very much!

coev/profiles.go
```diff
@@ -7,15 +7,15 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/veraison/corim/corim"
 	"github.com/veraison/corim/extensions"
-	"github.com/veraison/eat"
 )
 
 // ProfileManifest associates an EAT profile ID with a set of extensions. It allows
```
yogeshbdeshpande:
```suggestion
// ProfileManifest associates an CoRIM profile ID with a set of extensions. It allows
```

comid/domain_membership_triple.go
```diff
@@ -0,0 +1,131 @@
...
+	return &DomainMembershipTriples{}
+}
+
+// Add a triple to the DomainMembershipTriples.
+func (o *DomainMembershipTriples) Add(triple DomainMembershipTriple) *DomainMembershipTriples {
+	*o = append(*o, triple)
+	return o
+}
+
+// Valid retursn an error if DomainMembershipTriples is empty or conatains
```
yogeshbdeshpande:
```suggestion
// Valid returns an error if DomainMembershipTriples is empty or conatains
```

comid/rawvalue.go
```diff
@@ -1,138 +1,345 @@
...
+	ret, err := NewMaskedRawValue(val)
+	if err != nil {
+		panic(err)
+	}
+
+	return ret
+}
+
+// IRawValueFactory defines the signature for the factory functions that may be
+// registred using RegisterRawValueType to provide a new implementation of the
```
yogeshbdeshpande:
```suggestion
// registered using RegisterRawValueType to provide a new implementation of the
```

comid/rawvalue.go
```diff
@@ -1,138 +1,345 @@
...
+		panic(err)
 	}
+
+	return ret
 }
 
-// Equal confirms if the RawValue instances are equal
-func (o RawValue) Equal(r RawValue) bool {
-	return reflect.DeepEqual(o, r)
+// NewRawValueWithMask returns a pointer to a new RawValues with a masked
```
yogeshbdeshpande:
```suggestion
// NewRawValueWithMask returns a pointer to a new RawValue with a masked
```

comid/rawvalue.go
```diff
@@ -1,138 +1,345 @@
...
-			return false
-		}
+	return ret
+}
 
-		if len(*mask) != len(claim) {
-			log.Printf("RawValue:CompareAgainstReference: Error: mask length")
-			return false
-		}
+// RawValue models a $raw-value-type-choice.  For now, the only available type is bytes.
```
yogeshbdeshpande: Is this true, I think, we MUST say, Available type is tagged bytes and also a Tagged Value which is bytes with a mask?

comid/domain_membership_triple.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	}
+
+	return nil
+}
+
+// DomainMembershipTriples is a container of DomainMembershipTriple instances.
+type DomainMembershipTriples []DomainMembershipTriple
+
+// NewDomainMebershipTriples returns a new empty NewDomainMebershipTriples.
```
yogeshbdeshpande:
```suggestion
// NewDomainMebershipTriples returns a new empty DomainMebershipTriples.
```

comid/domain_membership_triple.go
```diff
@@ -0,0 +1,131 @@
...
+	return &DomainMembershipTriples{}
+}
+
+// Add a triple to the DomainMembershipTriples.
+func (o *DomainMembershipTriples) Add(triple DomainMembershipTriple) *DomainMembershipTriples {
+	*o = append(*o, triple)
+	return o
+}
+
+// Valid returns an error if DomainMembershipTriples is empty or conatains
```
yogeshbdeshpande:
```suggestion
// Valid returns an error if DomainMembershipTriples is empty or contains
```

comid/triples.go
```diff
@@ -321,6 +387,30 @@ func (o *Triples) AddDomainDependency(val *DomainDependencyTriple) *Triples {
 	return o
 }
 
+// AddDomainMembership appends a domain-dependency-triple to DomainDependencies.
```
yogeshbdeshpande:
```suggestion
// AddDomainMembership appends a domain-membership-triple to DomainMemberships.
```

comid/triples.go
```diff
@@ -321,6 +387,30 @@ func (o *Triples) AddDomainDependency(val *DomainDependencyTriple) *Triples {
...
+	if o != nil {
+		if o.DomainMemberships == nil {
+			o.DomainMemberships = new(DomainMembershipTriples)
+		}
+		*o.DomainMemberships = append(*o.DomainMemberships, *val)
+	}
+
+	return o
+}
+
```
yogeshbdeshpande, line 401:
```suggestion
// AddCoswidTriple appends a CoswidTriple to CoswidTriples 
```

comid/triples.go
```diff
@@ -333,3 +423,15 @@ func (o *Triples) AddCondEndorseSeries(val *CondEndorseSeriesTriple) *Triples {
 
 	return o
 }
+
```
yogeshbdeshpande, line 428:
```suggestion
// AddCondEndorsement appends a CondEndorseTriple to CondEndorsements 
```

corim/profile.go
```diff
@@ -0,0 +1,188 @@
...
+
+// IProfileValue is the interface implemented by all types whose instances can
+// be used as Profile values.
+type IProfileValue interface {
+	extensions.ITypeChoiceValue
+}
+
+// Profile is an identification of a CoRIM profile that defines which of the
+// optional parts of CoRIM are required, which are prohibited, and which
+// extension points are exercised and how.
```
yogeshbdeshpande:
```suggestion
// extension points are exercised and how. Currently supported types are URI and Tagged OID as per section 4.1.4 of CoRIM Specification
```

corim/profile.go
```diff
@@ -0,0 +1,188 @@
...
+// Valid returns an error if the underlying profile value is invalid
+func (o *Profile) Valid() error {
+	if o.IsNil() {
+		return errors.New("cannot be nil")
+	}
+
+	return o.Value.Valid()
+}
+
+// IsNil returns true if the *Profile is nil or if its innver value is nil.
```
yogeshbdeshpande:
```suggestion
// IsNil returns true if the *Profile is nil or if its inner value is nil.
```

corim/profile_handling.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Contributors to the Veraison project.
+// Copyright 2024-2026 Contributors to the Veraison project.
```
yogeshbdeshpande, line 1: One general request: Now that we have a profile.go that define how CoRIM Profiles (which are either Tagged OIDs or URI) needs to be instantiated, please change the name of this file to either 

`profile-extensions.go` or anything similar and more logical rather than just profiles.go 

Otherwise I got the impression similar to digest.go and digests.go ( which has []digest )

yogeshbdeshpande (commented):
Have left some nits for you to look at, in general LGTM!

Please correct these minor comments and then merge it!

yogeshbdeshpande (approved):
LGTM!

Thanks for incorporating my comments!

Issue #267, opened 2026-05-20
corim: kid within a signed corim


atulfj:
The issue is related to `kid` being absent within a signed corim that was created using `cocli`.

It led me to a [discussion thread about the significance of `kid` in an old corim draft](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/issues/363), where the conclusion seemed to be that `kid` should not be a mandatory field. The current corim CDDL reflects it in [protected-corim-header-map.cddl](https://github.com/ietf-rats-wg/draft-ietf-rats-corim/blob/main/cddl/protected-corim-header-map.cddl). However, [a signed corim in veraison/corim-rs contains the kid](https://github.com/veraison/corim-rs/blob/ec30440b4977bd7c9d355fcf15dd3347c866d4c2/src/corim.rs#L1967) whereas [the same structure in the Go-based library does not](https://github.com/veraison/corim/blob/99498ccd67775d44b326e9e64255af6d8129f4ba/corim/signedcorim.go#L27). What is the reason for this difference in implementations?

For added context, the corim verifier library, `cover`, first stores a key against its `kid` and then looks it up during verification. If a signed corim is created using `cocli` and passed to `cover`, it fails because of the empty `kid`.

setrofim:
> What is the reason for this difference in implementations?

These implementations were created at different points in time,  and with different use cases in mind. `corim-rs` includes `kid` support, because it was mandatory at the time it was initially developed. Go implementation was treating it as optional even while it wasn't, and chose to implement `x5chain` instead.  `cover` uses `kid` because it's simpler than dealing with x509 certs, and it was never intended to be anything more than a self-contained PoC to prove out section 9 of the spec.

In any case, adding support for `kid` in this implementation should be straight forward.

setrofim:
`kid` support has now been added in `corim` (https://github.com/veraison/corim/pull/269) and `cocli` (https://github.com/veraison/cocli/pull/55).

PR #269, opened 2026-05-20
feat: add support for kid COSE header

setrofim:
Add support for the key-id (label 4) COSE protected header. This is not required by the spec, but may be useful as an alternative to x5chain in some use cases where certified keys aren't used.

Addresses: https://github.com/veraison/corim/issues/267


yogeshbdeshpande (approved):
LGTM!

thomas-fossati (approved):
Thanks!

PR #284, opened 2026-09-01
feat(cotl): add Concise Tag List (CoTL) package

larrydewey:
Donate the veraison/cotl implementation upstream as corim/cotl, following the module's package-per-tag layout (comid, coserv, coev). Adapts the extension mechanism to the current extensions.Map/IMapValue API and wires the package into the Makefile and README package list.

### DIFF
--- a/cotl/cotl.go
+++ b/cotl/cotl.go
@@ -1,0 +1,259 @@
+// Copyright 2026 Contributors to the Veraison project.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package cotl implements the Concise Tag List (CoTL) data structure defined
+// in Section 6 of draft-ietf-rats-corim-11, including CBOR (tag 508) and JSON
+// serialization, validation, temporal appraisal checks, and Veraison-style
+// map extensions on the concise-tl-tag extension point.
+package cotl
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	cbor "github.com/fxamacker/cbor/v2"
+	"github.com/veraison/corim/corim"
+	"github.com/veraison/corim/encoding"
+	"github.com/veraison/corim/extensions"
+	"github.com/veraison/swid"
+)
+
+// CotlTag is the CBOR tag number for a tagged-concise-tl-tag
+// (draft-ietf-rats-corim-11, Section 6).
+const CotlTag uint64 = 508
+
+var (
+	em cbor.EncMode
+	dm cbor.DecMode
+)
+
+func init() {
+	encOpt := cbor.EncOptions{
+		Sort:          cbor.SortCoreDeterministic,
+		IndefLength:   cbor.IndefLengthForbidden,
+		NilContainers: cbor.NilContainerAsEmpty,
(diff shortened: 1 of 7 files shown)


cotl/cotl.go
```diff
@@ -0,0 +1,259 @@
...
+// input are surfaced as errors instead of crashing the calling process.
+func populateSafely(o *ConciseTlTag, content []byte) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("CBOR decoding failed: %v", r)
+		}
+	}()
+
+	// Gate on RFC 8949 well-formedness first: the corim encoding helpers do
+	// not defend against malformed map headers (they can loop indefinitely).
```
setrofim, line 208: This is something we need to fix inside encoding helpers. Could you elaborate on this/give example CBOR which causes `PopulateStrtuctFromCBOR()` to loop indefinitely?

setrofim: OK I think I see what the issue might be (I commented out the `Wellformed` check and re-ran your tests). I think the decoder is not  looping indefinitely (I've gone through the code and don't see where an infinite loop might occur -- let me know if missed something), but rather are trying to allocate massive buffers when it interprets random byte sequence as a massive map header. 

Th fix for that would then be to do the well-formed check on the entire input before attempting to allocate anything. So the check below should probably be done as part of `PopulateStructFromCBOR` rather than here.

thomas-fossati: An alternative could be to configure the decoder with stricter limits on the acceptable map's shape/size, using `MaxMapPairs` and `MaxNestedLevels`.

setrofim: I've added that here: https://github.com/veraison/corim/pull/285

That should remove the need for this check here.

setrofim: > An alternative could be to configure the decoder with stricter limits on the acceptable map's shape/size, using MaxMapPairs and MaxNestedLevels.

We could, but whatever limit we come up with will be essentially arbitrary, and would hypothetically prevent valid large inputs from being parsed.  Besides, having the well-formed check prior to decoding is good for robustness anyway.

setrofim (approved):
LGTM!

cotl/cotl.go
```diff
@@ -0,0 +1,259 @@
...
+		TimeTag:     cbor.DecTagOptional,
+	}.DecMode()
+	if err != nil {
+		panic(err)
+	}
+	dm = d
+}
+
+// TagIdentityMap identifies a CoMID or CoSWID tag, either as the identity of
+// the CoTL itself or as an entry in its tags-list.
```
yogeshbdeshpande:
```suggestion
// TagIdentityMap identifies a CoMID or CoSWID tag, as an entry in its tags-list or 
// as the identity of the CoTL itself
```

yogeshbdeshpande (approved):
LGTM!,

thomas-fossati (approved):
🚢 it!

yogeshbdeshpande:
Thank you so much @larrydewey for this addition.

PR #285, opened 2026-09-02
fix(encoding): ensure input is well-formed before allocation

setrofim:
Perform `dm.Wellformed()` check on the input at the beginning of `structFieldsCBOR.FromCBOR()`. This prevents, among other things, random byte sequences from being interpreted as map headers with a large size and then waiting on the allocation.

Thanks to @larrydewey  for identifying the issue.

### DIFF
--- a/encoding/cbor_test.go
+++ b/encoding/cbor_test.go
@@ -279,17 +279,30 @@ func Test_structFieldsCBOR_CBOR_decode_indefinite(t *testing.T) {
 	assert.Equal(t, []int{0, 1, 2, 3, 4}, sfOut.Keys)
 }
 
+func Test_structFieldsCBOR_CBOR_decode_malformed(t *testing.T) {
+	data := []byte{
+		0xcb, 0xba, 0x5a, 0x47, 0x42, 0xe0, 0xa5, 0x4e,
+	}
+
+	dm, err := cbor.DecOptions{}.DecMode()
+	require.NoError(t, err)
+
+	sfOut := newStructFieldsCBOR()
+	err = sfOut.FromCBOR(dm, data)
+	assert.ErrorContains(t, err, "malformed CBOR")
+}
+
 func Test_structFieldsCBOR_CBOR_decode_negative(t *testing.T) {
 	dm, err := cbor.DecOptions{}.DecMode()
 	require.NoError(t, err)
 
 	sfOut := newStructFieldsCBOR()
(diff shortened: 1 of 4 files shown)


larrydewey (approved):

encoding/cbor_test.go
```diff
@@ -279,17 +280,33 @@ func Test_structFieldsCBOR_CBOR_decode_indefinite(t *testing.T) {
...
+	}
+
+	dm, err := cbor.DecOptions{}.DecMode()
+	require.NoError(t, err)
+
+	sfOut := newStructFieldsCBOR()
+	start := time.Now()
+	err = sfOut.FromCBOR(dm, data)
+	duration := time.Since(start)
+	assert.Less(t, duration.Seconds(), 1.0)
```
thomas-fossati: is this really meaningful?

setrofim: Without the fix, the input results in a large allocation (on my machine takes several seconds to resolve).

thomas-fossati: We already have the "malformed" signal.
I believe checking that is sufficient to ensure the well-formedness has been verified.

setrofim: But without the fix, it  is also reported as  malformed (albeit with a somewhat different message). The fix just make the detection faster. 

We're not detecting malformed inputs we weren't detecting before. We're just detecting them much faster (because we verify the entire input before attempting allocation). The issue being fixed is temporal, so the test needs to reflect that.

thomas-fossati: Understood, but whichever time bound we choose (1s, 1ms, etc.) will necessarily depend on the platform on which the test is executed rather than being an intrinsic property of the SUT.
What matters is that the error is identified through the expected code path, not how much it takes before the condition is evaluated.

setrofim: Sure, the specific timing will be platform-specific, but the point here is the orders of magnitude difference between fix and no fix, which can be observed on any platform. However, I'm fine with removing the timing from the test and just leaving it based solely on matching the expected message.

setrofim:
Now that https://github.com/veraison/corim/pull/284 has been merged, I rebased on top and removed the `dm.Wellformed()` check from `cotl` as it is now being done inside `PopulateStructFromCBOR()`.

yogeshbdeshpande (approved):
LGTM!

thomas-fossati (approved):
🚢 it, thanks!

